PR #19, opened 2022-05-13
Enable the app deployment from a private container registry 

### LINKED ISSUE #18: Enable the app deployment from a private container registry
Regarding streams app deployments, the manager is limited to open (public) container repositories. In order to allow users to deploy an app from a private repo, a possibility to add imagePullSecrets through quick and quick-cli should be added.
Example: 

```
 quick app deploy tiny-url-counter \
     --registry bakdata \
     --image quick-examples-tinyurl \
     --tag 0.0.1 \
     --imagePullSecrets=secret\
     --args input-topics=track-fetch output-topic=count-fetch productive=false
```

We assume that a user has (uses) only one secret. Checking multiple secrets won’t be supported for now.

DawidNiezgodka:
This pull request resolves #18.

Assumptions:
1. A user provides only a single secret,
2. We don't make any secret or secret-name validity check, i.e. when a user inputs an empty key name or two secret (comma separated), then the app deployment simply fails.

### DIFF
--- a/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
+++ b/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
@@ -48,5 +48,6 @@ spec:
             requests:
               memory: [(${resourceConfig.memory.request})]
               cpu: [(${resourceConfig.cpu.request})]
-      imagePullSecrets: [# th:if="${imagePullSecret}"]
+      [# th:if="${imagePullSecret}"]
+      imagePullSecrets:
         - name: [(${imagePullSecret})] [/]
--- a/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
+++ b/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
@@ -51,3 +51,4 @@ spec:
       [# th:if="${imagePullSecret}"]
       imagePullSecrets:
         - name: [(${imagePullSecret})] [/]
+
--- a/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
+++ b/manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
@@ -48,3 +48,5 @@ spec:
             requests:
               memory: [(${resourceConfig.memory.request})]
               cpu: [(${resourceConfig.cpu.request})]
+      imagePullSecrets: [# th:if="${imagePullSecret}"]
+        - name: [(${imagePullSecret})] [/]
--- a/manager/src/test/java/com/bakdata/quick/manager/application/resources/ApplicationResourceLoaderTest.java
+++ b/manager/src/test/java/com/bakdata/quick/manager/application/resources/ApplicationResourceLoaderTest.java
@@ -27,13 +27,7 @@ import com.bakdata.quick.manager.k8s.KubernetesResources;
 import com.bakdata.quick.manager.k8s.KubernetesTest;
 import com.bakdata.quick.manager.k8s.resource.QuickResources.ResourcePrefix;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import io.fabric8.kubernetes.api.model.Container;
-import io.fabric8.kubernetes.api.model.EnvFromSource;
-import io.fabric8.kubernetes.api.model.EnvVar;
-import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.fabric8.kubernetes.api.model.IntOrString;
-import io.fabric8.kubernetes.api.model.PodSpec;
-import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.*;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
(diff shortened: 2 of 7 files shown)


manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
```diff
@@ -48,3 +48,5 @@ spec:
             requests:
               memory: [(${resourceConfig.memory.request})]
               cpu: [(${resourceConfig.cpu.request})]
+      imagePullSecrets: [# th:if="${imagePullSecret}"]
+        - name: [(${imagePullSecret})] [/]
```
raminqaf, line 52: newline

manager/src/test/java/com/bakdata/quick/manager/application/resources/ApplicationResourceLoaderTest.java
```diff
@@ -27,13 +27,7 @@
...
 import com.bakdata.quick.manager.k8s.resource.QuickResources.ResourcePrefix;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import io.fabric8.kubernetes.api.model.Container;
-import io.fabric8.kubernetes.api.model.EnvFromSource;
-import io.fabric8.kubernetes.api.model.EnvVar;
-import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.fabric8.kubernetes.api.model.IntOrString;
-import io.fabric8.kubernetes.api.model.PodSpec;
-import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.*;
```
raminqaf, line 30: please avoid wildcard imports

raminqaf (changes requested):

manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
torbsto, line 52: New line please. There is also a corresponding IntelliJ feature that you can activate in the settings

manager/src/test/java/com/bakdata/quick/manager/application/resources/ApplicationResourceLoaderTest.java
torbsto, line 30: Please don't use star imports. I thought this is covered by the code style. Maybe it's also a different setting in IntelliJ

manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
```diff
@@ -48,3 +48,5 @@ spec:
             requests:
               memory: [(${resourceConfig.memory.request})]
               cpu: [(${resourceConfig.cpu.request})]
+      imagePullSecrets: [# th:if="${imagePullSecret}"]
```
torbsto, line 51: I think I'd prefer something like this:
```suggestion
      [# th:if="${imagePullSecret}"]
      imagePullSecrets: 
```

manager/src/test/java/com/bakdata/quick/manager/application/resources/ApplicationResourceLoaderTest.java
raminqaf: This is also an IntelliJ configuration. https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports-

raminqaf:
Please fix the failing test `shouldReturnComplexType`

torbsto:
The test is not related to this PR, but seems to be flaky on the CI runners

DawidNiezgodka:
> Please fix the failing test `shouldReturnComplexType`

I was confused why it failed as it seems to have nothing to do with this PR.

manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
DawidNiezgodka: If we make it this way, then secret=null leads to an empty ArrayList in the field imagePullSecrets.
In the original version, we have imagePullSecrets=null.
Do we really want to have an empty array here or null?

torbsto: Empty list seems to be the better choice then I think

manager/src/main/resources/k8s/templates/streamsApp/deployment.th.yaml
DawidNiezgodka: Hmm even though I put it there and change the corresponding setting in intellij, it removes the extra newline

DawidNiezgodka: ok nvm i added it manually

torbsto (commented):
Looking good! Please also update the [OpenAPI scheme that describes the `ApplicationCreationData`](https://github.com/bakdata/quick/blob/f4c49c3721e9806baa93a5b1723eaa09717c284a/openapi/spec/schemas.yaml#L116). We need this for updating the python client.

torbsto (approved):

raminqaf (approved):
LGTM

PR #27, opened 2022-05-17
Add configuration for schema format

### LINKED ISSUE #21: New configuration variable for schema type
As described in #17, we require a configuration variable for letting users set their desired schema format (e.g., Avro, Protobuf).
- [ ] Java Configuration class
- [ ] Helm chart support

torbsto:
Closes #21 

This adds a configuration `quick.schema.format` that can be set to Avro or Protobuf. 
It also:
- moves the `AvroConfig` from `quick.avro` to `quick.schema.avro` to align it with the new class (**breaking change**)
- adds utility for improved testing of different configurations
- a condition for loading beans only if Avro is enabled


common/src/testFixtures/java/com/bakdata/quick/common/ConfigUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.bakdata.quick.common;
+
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.context.env.MapPropertySource;
+import io.micronaut.context.env.PropertySource;
+import java.util.Map;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
```
raminqaf, line 25: Why use this annotation if you can make everything static? I think making the class `final` and the members `static` would be easier to read and understand.

docs/docs/user/reference/configuration.md
```diff
@@ -17,6 +17,14 @@
 | `QUICK_MIRROR_PREFIX` | :material-close: | Prefix of Kubernetes deployments for mirror deployments |
 | `QUICK_MIRROR_PATH`   | :material-close: | REST subpath where the mirror's API is running          |
 
+
+## Schema
+
+| Environment Variable          | Required                             | Description                                                            |
+|-------------------------------|--------------------------------------|------------------------------------------------------------------------|
+| `QUICK_SCHEMA_FORMAT`         | :material-close:                     | Schema format Quick should use. Valid values: Avro (default), Protobuf |
```
raminqaf, line 25: Perhaps address that the valid values are not case sensitive

raminqaf (approved):
LGTM just some tiny points that I noticed

docs/docs/user/reference/configuration.md
torbsto: I'd expect it to be case-insensitive, therefore explicitly mentioning it seems unnecessary to me.

PR #29, opened 2022-05-25
Use ParsedSchema instead of Schema

### LINKED ISSUE #24: `TopicTypeService` should be aware of different schema formats
In the implementation of `TopicTypeService` (mainly `QuickTopicTypeService`), Quick configures the SerDe (see #1) and TypeResolver (see #2). Therefore, Quick should be aware of Protobuf here to appropriately configure those elements.

torbsto:
Closes #24 

With this PR, Quick uses `ParsedSchema` instead of Avro's `Schema` to represent schemas. The former allows for different implementation, i.e., Avro and Protobuf. In a follow-up PR, I'll add the `ProtobufProvider` for the schema registry client.
I've also deleted the `InferTopicTypeService` since it's not used anymore.


raminqaf (approved):
LGTM!

PR #31, opened 2022-06-02
Convert GraphQL schema to Protobuf

### LINKED ISSUE #4: Convert Protobuf to GraphQL
We require the automatic conversion from the Kafka schema to GraphQL for checking schema compatibility and automatically creating the target schema.
With #3, we therefore need the conversion from Protobuf to GraphQL.

raminqaf:
Closes #4 
This pull request adds the logic that converts GraphQL schemas to Protobuf


manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,310 @@
...
+        currentMessage.addField(buildFieldWithType(graphQLFieldDefinition.getName(),
+            fieldNumber,
+            Type.TYPE_MESSAGE,
+            graphQLObjectType.getName(),
+            label));
+
+        final Builder descriptorProtoBuilder = DescriptorProto.newBuilder();
+        descriptorProtoBuilder.setName(graphQLObjectType.getName());
+
+        final FileDescriptorProto.Builder tempFile = FileDescriptorProto.newBuilder();
```
torbsto, line 203: I don't see why you would need a tempFile. You can move the check to each corresponding addition

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,310 @@
...
+    /**
+     * This function creates a FieldDescriptorProto object for a specific type. These fields could be either a type of
+     * TYPE_MESSAGE or TYPE_ENUM. If a FieldDescriptorProto object type is one of these the property TypeName must be
+     * specified otherwise a {@link DescriptorValidationException} is thrown.
+     *
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type#TYPE_MESSAGE
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type#TYPE_ENUM
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Builder#setTypeName(String)
+     */
+    @NotNull
```
torbsto, line 221: No need for `@NotNull`, we mark `@Nullable`

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /**
+     * This function creates a {@link EnumDescriptorProto} object from a scalar GraphQL type.
+     */
+    private static EnumDescriptorProto createEnumDescriptor(final GraphQLEnumType enumType) {
+        final List<EnumValueDescriptorProto> values = enumType.getDefinition()
+            .getEnumValueDefinitions()
+            .stream()
+            .map(enumValueDefinition -> EnumValueDescriptorProto.newBuilder()
```
torbsto, line 288: I've just noticed that this is a bit tricky. Given the following GraphQL
```graphql
enum Foo {
  A, B, C
}

enum Bar {
  A, B, C
}
```
the resulting protobuf would be invalid, because the enum values would clash in the namespace.
You could prefix them with the type name.

manager/src/test/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverterTest.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.jupiter.api.TestInfo;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class GraphQLToProtobufConverterTest {
```
torbsto, line 35: :heart_eyes:

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,310 @@
...
+
+    private static final Map<GraphQLScalarType, Type> SCALAR_MAPPING = scalarTypeMap();
+
+    @Inject
+    public GraphQLToProtobufConverter(final ProtobufConfig protobufConfig) {
+        this(protobufConfig.getProtobufPackage());
+    }
+
+    @VisibleForTesting
+    public GraphQLToProtobufConverter(final String protobufPackage) {
```
torbsto, line 77: Well, you also could just use `new GraphQLToProobufConverter(new ProtobufConfig(...))`

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLConverter.java
```diff
@@ -0,0 +1,59 @@
...
+import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLSchema;
+import graphql.schema.idl.RuntimeWiring;
+import graphql.schema.idl.SchemaGenerator;
+import graphql.schema.idl.SchemaParser;
+import graphql.schema.idl.TypeDefinitionRegistry;
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+
+/**
+ * An interface for converting GraphQL schema to a {@link ParsedSchema} object.
```
torbsto, line 30:
```suggestion
 * Converter for transforming a GraphQL schema to a {@link ParsedSchema} object.
```
Just as an example, similar to the comment above

torbsto (commented):
Thanks for the PR! Looks good overall, just a couple of nits and one open question regarding the conversion of enums.

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,310 @@
...
+        this.protobufPackage = protobufPackage;
+    }
+
+    @Override
+    public ParsedSchema convert(final String graphQLSchema) {
+        try {
+            final Descriptor descriptor = this.convertToDescriptor(graphQLSchema);
+            return new ProtobufSchema(descriptor);
+        } catch (final DescriptorValidationException exception) {
+            throw new BadRequestException(
```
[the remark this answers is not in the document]
raminqaf: I am a bit confused about this comment. You are using the same exception in the Avro converter here: https://github.com/bakdata/quick/blob/ed00e6c8f164d66110030dd2cd9b13a5c5015dd1/manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToAvroConverter.java#L160

raminqaf:
@torbsto Thanks for the review! I applied them to my code. here are some notes:
- Based on the [style guide of buf](https://docs.buf.build/lint/rules#enum_value_prefix), I prefixed all the enum fields with the type name. Moreover, I added an `UNSPECIFIED` field to the beginning of the enum descriptor.

- I refactored the Protobuf conversion tests, and the test style is unified with the tests in the Avro conversion.

- I removed the `tempFile` builder. I check if a message/enum exists before adding it to the file.

- I applied the Javadoc style you wished for and will keep this style in my future PRs.

- I replaced the `BadArgumentException` with an `InternalErrorException`, since the `DescriptorValidationErrorException` accours when do the conversion wrong.

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,326 @@
...
+     * @param typeName prefix to the field name
+     * @param enumValueDefinitionName the graphQL enum field name
+     * @return enum field name
+     */
+    private static String generateEnumFieldName(final String typeName, final String enumValueDefinitionName) {
+        return typeName.toUpperCase() + "_" + enumValueDefinitionName;
+    }
+
+    private static Map<GraphQLScalarType, Type> scalarTypeMap() {
+        return Map.ofEntries(
```
torbsto, line 315: Any particular reason you're using Map.ofEntries instead of Map.of?

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,326 @@
...
+    /**
+     * Creates a FieldDescriptorProto object for a specific type. These fields could be either a type of
+     * TYPE_MESSAGE or TYPE_ENUM. If a FieldDescriptorProto object type is one of these the property TypeName must be
+     * specified otherwise a {@link DescriptorValidationException} is thrown.
+     *
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type#TYPE_MESSAGE
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type#TYPE_ENUM
+     * @see com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Builder#setTypeName(String)
+     */
+    private static FieldDescriptorProto buildFieldWithType(
```
torbsto, line 208:
```suggestion
    private static FieldDescriptorProto createFieldWithType(
```

All the methods returning a proto element start with `create...`, any reason this doesn't?

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,326 @@
...
+                Label.LABEL_REPEATED);
+        } else {
+            throw new BadArgumentException(
+                String.format("Type %s not recognized", GraphQLTypeUtil.simplePrint(graphQLType)));
+        }
+    }
+
+    private static void handleObjectType(
+        final FileDescriptorProto.Builder fileBuilder,
+        final Builder currentMessage,
```
torbsto, line 182:
```suggestion
        final Descriptor.Builder currentMessage,
```
As a parameter, `Builder` is to generic

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,326 @@
...
+                graphQLList.getWrappedType(),
+                fieldNumber,
+                Label.LABEL_REPEATED);
+        } else {
+            throw new BadArgumentException(
+                String.format("Type %s not recognized", GraphQLTypeUtil.simplePrint(graphQLType)));
+        }
+    }
+
+    private static void handleObjectType(
```
torbsto, line 180: Could you add JavaDoc here as well?

torbsto (commented):
I suggest we document what GraphQL schema elements Quick supports, both from a query and conversion perspective. I *think* unions and interfaces are the only elements Quick doesn't support. This can be done in a follow-up PR though

common/src/main/java/com/bakdata/quick/common/config/ProtobufConfig.java
```diff
@@ -0,0 +1,57 @@
...
+     * Default constructor.
+     *
+     * @param protobufPackage Holds the name of the protobuf package where the object is stored.
+     */
+    @ConfigurationInject
+    public ProtobufConfig(final String protobufPackage) {
+        if (!NAMESPACE_PATTERN.matcher(protobufPackage).matches()) {
+            throw new ConfigurationException(
+                String.format(
+                    "The Protobuf package %s does not fulfill the naming convention of Protobuf specification.",
```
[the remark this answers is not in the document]
raminqaf: Not from Google directly but as a style guide from buf.

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
```diff
@@ -0,0 +1,326 @@
...
+
+    /**
+     * Generates the enum field name by prefixing the uppercase type name to the field name.
+     *
+     * @param typeName prefix to the field name
+     * @param enumValueDefinitionName the graphQL enum field name
+     * @return enum field name
+     */
+    private static String generateEnumFieldName(final String typeName, final String enumValueDefinitionName) {
+        return typeName.toUpperCase() + "_" + enumValueDefinitionName;
```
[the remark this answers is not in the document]
raminqaf: I think it is not a bad idea

common/src/main/java/com/bakdata/quick/common/config/ProtobufConfig.java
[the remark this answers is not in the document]
torbsto: I think we shouldn't enforce a style guide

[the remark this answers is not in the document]
raminqaf: Should I remove it then?

common/src/main/java/com/bakdata/quick/common/config/ProtobufConfig.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Getter;
+
+/**
+ * Configuration regarding the converted GraphQL to Protobuf schemas.
+ *
+ */
+@ConfigurationProperties(ProtobufConfig.PREFIX)
+@Getter
+public class ProtobufConfig {
+    private static final String PROTO_FILE_EXTENSION = ".proto";
```
torbsto, line 30: It's just a coincidence though that the config suffix is the same as the proto file extension. To me, it's a bit confusing to have the extension variable here used for the prefix

common/src/main/java/com/bakdata/quick/common/config/ProtobufConfig.java
```diff
@@ -0,0 +1,43 @@
...
+import io.micronaut.context.annotation.ConfigurationProperties;
+import lombok.Getter;
+
+/**
+ * Configuration regarding the converted GraphQL to Protobuf schemas.
+ *
+ */
+@ConfigurationProperties(ProtobufConfig.PREFIX)
+@Getter
+public class ProtobufConfig {
```
torbsto, line 29: I think there is no test for this class yet

raminqaf:
The only thing that bothers me is the `protobufPackage` in the `ProtobufConfig` class. I cannot change it to only `package` because it is a keyword in Java. So the config would look like this:
`quick.schema.protobuf.protobuf-package`. How about we change it to `packageName`?

torbsto:
You could do something like this:

```java
@ConfigurationInject
public ProtobufConfig(@Property(name = PREFIX + ".package") final String protobufPackage) {
    this.protobufPackage = protobufPackage;
}
```

But I'm also fine with `packageName`

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLConverter.java
```diff
@@ -0,0 +1,59 @@
...
+     * do the conversion.
+     *
+     * @param schema The string containing the GraphQL schema.
+     * @return a {@link GraphQLObjectType} object, which contains information of the root object in the schema.
+     */
+    default GraphQLObjectType getRootTypeFromSchema(String schema) {
+        // extending the schema with an empty query type is necessary because parsing fails otherwise
+        final SchemaParser schemaParser = new SchemaParser();
+
+        // extending the schema with an empty query type is necessary because parsing fails otherwise
```
torbsto, line 47: Comment is duplicated, see line 44

manager/src/test/resources/schema/graphql/shouldConvertListType.graphql
```diff
@@ -0,0 +1,11 @@
+type ListType {
+    optionalSimpleList: [Int]
+    optionalComplexList: [ComplexObject]
+
+    requiredSimpleList: [Int]!
+    requiredComplexList: [ComplexObject]!
```
torbsto, line 6: Can you add a test for `[Object!]` and `[Object!]!`?

torbsto, line 6: Can you add a test for `[Object!]` and `[Object!]!`?

manager/src/main/java/com/bakdata/quick/manager/graphql/GraphQLToProtobufConverter.java
torbsto: This is not resolved yet

torbsto: This is not resolved yet

torbsto (approved):

PR #33, opened 2022-06-09
Changes in StreamsStateController (prev. ApplicationResource) that enable testing thereof

DawidNiezgodka:
The existing code base has been slightly adjusted to enable testing of the StreamsStateController (previously named ApplicationResource). In order to achieve this, a custom provider that imitates (and extends) the functionality of the javax.inject.Provider has been added. This enables a necessary provision of information (via dependency injection) about the streams app state to the controller, and thus facilitates (enables) testing.  
The test shows that the controller is capable of retrieving a mapping between partitions and ports in a correct way.

Unfortunately this detour does not solve the main problem which concerns passing information between the application and the controllers. The possible reason behind this is the necessity to pass info that involves generic types, which doesn't play out well with dependency injection.


DawidNiezgodka:
Oddly enough the first integration test fails after the introduction of these changes. I'm working on it.

mirror/src/main/java/com/bakdata/quick/mirror/StreamsStateController.java
```diff
@@ -17,29 +17,32 @@
...
+import org.jetbrains.annotations.NotNull;
 import org.jooq.lambda.Seq;
+import javax.inject.Inject;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * REST API exposing current Kafka Streams state.
  */
 @Controller("/mirror")
```
torbsto, line 37: While we're at it, what do you think about also changing this? Maybe to `streams`?

mirror/src/main/java/com/bakdata/quick/mirror/MirrorApplication.java
```diff
@@ -31,17 +31,21 @@
 import com.bakdata.quick.common.util.CliArgHandler;
 import com.bakdata.quick.mirror.base.HostConfig;
 import com.bakdata.quick.mirror.base.QuickTopologyData;
+import com.bakdata.quick.mirror.service.QueryContextProvider;
 import com.bakdata.quick.mirror.service.QueryServiceContext;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import io.micronaut.configuration.picocli.MicronautFactory;
 import io.micronaut.context.ApplicationContext;
+import io.micronaut.context.ProviderFactory;
```
torbsto, line 39: There are a lot of unused imports in this PR. Could you remove them?

PR #36, opened 2022-06-15
Reduce responsibilities of TypeResolver

torbsto:
Before this commit, the TypeResolver's interface comprised:
- fromString
- fromObject
- toJSONString
- elementType

This is now reduced to only fromString.

fromObject was completely removed, toJsonString is done by Jackson. The elementType interface was required for deserialization with Jackson. Since this is equivalent to TypeResolver's fromString, it was replaced with it.


common/src/test/java/com/bakdata/quick/common/api/client/TopicDataClientTest.java
```diff
@@ -44,7 +45,7 @@ class TopicDataClientTest {
     private final String host = String.format("%s:%d", this.server.getHostName(), this.server.getPort());
     private final MirrorHost mirrorHost = new MirrorHost(this.host, MirrorConfig.directAccess());
     private final MirrorClient<String, TopicData> topicDataClient =
-        new DefaultMirrorClient<>(this.mirrorHost, this.client, this.mapper.constructType(TopicData.class));
+        new DefaultMirrorClient<>(this.mirrorHost, this.client, new KnownTypeResolver<>(TopicData.class, this.mapper));
```
raminqaf, line 48: How about we change the name of this test? It was pretty misleading the first time I looked at it. For example `TopicRegistryMirrorClientTest` or similar to that.

common/src/main/java/com/bakdata/quick/common/type/QuickTopicType.java
```diff
@@ -22,39 +22,110 @@
...
-
-    QuickTopicType(final Supplier<Serde<?>> serdeSupplier, final Supplier<TypeResolver<?>> resolverSupplier) {
-        this.serdeSupplier = serdeSupplier;
-        this.resolverSupplier = resolverSupplier;
-    }
+    SCHEMA {
+        @Override
+        public <K> TypeResolver<K> getTypeResolver(@Nullable final ParsedSchema parsedSchema) {
+            Objects.requireNonNull(parsedSchema, "Schema must not be null for Avro types");
+            if (!(parsedSchema instanceof AvroSchema)) {
```
raminqaf, line 43: I guess later on we add another if-else for Protobuf support. Correct?

raminqaf (approved):
LGTM for the refactoring part! I had some small points and questions that I wrote in my reviews.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorClient.java
```diff
@@ -45,47 +43,45 @@
...
     private final HttpClient client;
-    private final JavaType valueType;
-    private final JavaType listType;
+    private final MirrorValueParser<V> parser;
 
     /**
-     * Constructor for client.
+     * Constructor for the client.
      *
-     * @param topicName name of the topic the mirror is deployed
```
DawidNiezgodka, line 54: Maybe we'll add a preposition here, for example under which or for which the mirror is deployed

DawidNiezgodka (approved):

common/src/main/java/com/bakdata/quick/common/resolver/GenericAvroResolver.java
```diff
@@ -38,38 +29,12 @@
 public class GenericAvroResolver implements TypeResolver<GenericRecord> {
 
     private final JsonAvroConverter converter;
-    private final JavaType elementType;
     @Nullable
     private Schema schema = null;
```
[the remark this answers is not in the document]
torbsto: Oh, true. I think this should even be final and not nullable anymore

common/src/main/java/com/bakdata/quick/common/type/QuickTopicType.java
torbsto: Sort of. My idea is to deprecate `SCHEMA` and instead add `AVRO` and `PROTO`

common/src/main/java/com/bakdata/quick/common/resolver/KnownTypeResolver.java
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.UncheckedIOException;
+
+/**
+ * Type resolver for known types.
+ *
+ * @param <T> type to resolve
+ */
+public class KnownTypeResolver<T> implements TypeResolver<T> {
```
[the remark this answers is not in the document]
torbsto: Added some context

PR #38, opened 2022-06-28
Introduces routing based on partition-host mapping

### LINKED ISSUE #35: Partition-mapping routing in the gateway to avoid selecting a wrong replica in a multi-replica scenario
When we have a single mirror, then the situation is clear. All partitions are located on this particular mirror, and when there is a request for the value of a given key, it is retrieved seamlessly. 
The problem arises when we have more than one replica of a mirror:

**Scenario**: _Mirror1_ stores partitions 1 and 4, and _Mirror2_ stores partitions 2 and 3. Let's say that we want to get data for the `key="x"`. Our hashing function h says that `h("x") = 1 `which means that the value for the key "x" is stored in the first partition, which is located in _Mirror1_. 
Currently, a request from gateway goes to the (Kubernetes?) service that chooses a replica in a round-robin fashion. This means that it is statistically half of the time wrong. When this happens, the request has to be redirected to the other replica.

Because of the fact that a mirror knows which partitions it stores (it has a mapping between a partitions and a host), we can use this information to introduce routing based on the partition mapping.

Questions:
1. How do we expose this info to the gateway?
2. In which part of the gateway should we put this functionality?

Extra ideas:
1. Caching
2. Mirror could announce itself by the gateway (observer pattern?)

DawidNiezgodka:
This PR introduces routing based on partition mapping, and thus resolves #35.

New / changed things:
1. `PartitionRouter<K>` that implements `Router<K>` - responsible for calling an end-point to fetch info about partition-host mapping and serving requests for a specific key (`getHost(Key k)`),
2. `PartitionFinder` and its implementation `DefaultPartitionFinder` that provides strategy for finding partitions,
3. For correct functioning of the router (key serialization), a specific Serde<K> is needed. Thus we have to provide it from classes that use DefaultMirrorClient. This clashes a little bit with the changes introduces by #36, where in many cases Mirror works only with TypeResolver. Thats why there was a need to create a wrapper over TypeResolver and KeySerde to be able to pass them together: `KeySerdeValResolverWrapper<K, V>`

This is a draft as there are no tests yet.

The logic for dealing with a dynamic addition of replicas a la change listener will be handled in a separate issue.


DawidNiezgodka:
I'm wondering if the router is not coupled too tight with the mirror. We already have a problem with a test in TopicDataClientTest because the router needs information from an endpoint before any mock values are provided.
```
   public DefaultMirrorClient(final String topicName, final MirrorHost mirrorHost,
                               final HttpClient client, final Serde<K> keySerde, TypeResolver<V> valueResolver) {
        this.client = client;
        this.parser = new MirrorValueParser<>(valueResolver, client.objectMapper());
        final StreamsStateHost streamsStateHost = StreamsStateHost.fromMirrorHost(mirrorHost);
        **this.router = new PartitionRouter<>(client, streamsStateHost, keySerde, topicName);**
        this.knownHosts = this.router.getAllHosts();
    }
```

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Fetches host-partition mapping from the StreamsStateController and updates the content
+     * of the partitionToHost map with the retrieved information
+     */
+    private void init() {
+        log.info("Initializing partition router...");
+        final String url = this.streamsStateHost.getPartitionToHostUrl();
+        final Request request = new Request.Builder().url(url).get().build();
+        makeAndProcessRequestForPartitionHostMapping(request);
```
torbsto, line 63: This (and the consequent methods) is basically an implementation of `DefaultMirrorClient`, isn't it? I'd be great if we could reuse it here, though I'm aware that it might be a bit tricky, because this would create a cyclic dependency right now. 
I think the way to go here is to have `BaseMirrorClient`, which is basically the `DefaultMirrorClient` from before. Additionally, you'd add a `PartitionedMirrorClient` which uses a `PartitionRouter` and the `BaseMirrorClient` to implement the logic of this PR.

torbsto (commented):
Could you make sure to set up the code style and inspections in IntelliJ?

torbsto:
For the merge conflicts, see https://bakdata.github.io/quick/0.6/developer/contributing/#dependent-pull-requests

torbsto:
> I'm wondering if the router is not coupled too tight with the mirror. We already have a problem with a test in TopicDataClientTest because the router needs information from an endpoint before any mock values are provided.
> 
> ```
>    public DefaultMirrorClient(final String topicName, final MirrorHost mirrorHost,
>                                final HttpClient client, final Serde<K> keySerde, TypeResolver<V> valueResolver) {
>         this.client = client;
>         this.parser = new MirrorValueParser<>(valueResolver, client.objectMapper());
>         final StreamsStateHost streamsStateHost = StreamsStateHost.fromMirrorHost(mirrorHost);
>         **this.router = new PartitionRouter<>(client, streamsStateHost, keySerde, topicName);**
>         this.knownHosts = this.router.getAllHosts();
>     }
> ```

You could use a supplier for providing the `PartitionRouter`

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,144 @@
...
+     * @param keySerde serializer for the key
+     * @param topic the name of the corresponding topic
+     */
+    public PartitionRouter(final HttpClient client, final StreamsStateHost streamsStateHost,
+                           final Serde<K> keySerde, final String topic) {
+        this.client = client;
+        this.streamsStateHost = streamsStateHost;
+        this.keySerde = keySerde;
+        this.topic = topic;
+        this.partitionFinder = MirrorConfig.getDefaultPartitionFinder();
```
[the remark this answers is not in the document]
DawidNiezgodka: Ok. I also added a new constructor in which you can inject PartitionerFinder.

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRegistryClient.java
```diff
@@ -26,10 +26,10 @@
 import io.reactivex.Completable;
 import io.reactivex.Flowable;
 import io.reactivex.Single;
+import lombok.extern.slf4j.Slf4j;
```
torbsto, line 29: There is still something wrong with your setup, because it reorders the imports in a wrong way

common/src/main/java/com/bakdata/quick/common/api/client/BaseMirrorClient.java
```diff
@@ -0,0 +1,129 @@
...
+ *    limitations under the License.
+ */
+
+package com.bakdata.quick.common.api.client;
+
+import com.bakdata.quick.common.api.model.mirror.MirrorHost;
+import com.bakdata.quick.common.api.model.mirror.MirrorValue;
+import com.bakdata.quick.common.exception.MirrorException;
+import com.bakdata.quick.common.resolver.TypeResolver;
+import org.jetbrains.annotations.Nullable;
```
torbsto, line 23: Please use `edu.umd.cs.findbugs.annotations.Nullable`

common/src/main/java/com/bakdata/quick/common/api/client/MirrorClient.java
```diff
@@ -58,4 +59,24 @@
      * @return True/False if key exists in mirror topic
      */
     boolean exists(final K key);
+
```
torbsto, line 62: Those two methods are only required for the base class? If so, I think it's best to split this into two separate interfaces

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private Map<Integer, String> makeRequestForPartitionHostMapping() {
+        final String url = this.streamsStateHost.getPartitionToHostUrl();
+        try (final ResponseBody responseBody = makeRequest(url)) {
+            final TypeReference<Map<Integer, String>> typeRef = new TypeReference<>() {};
+            final Map<Integer, String> partitionHostMappingResponse = this.client.objectMapper().readValue(
+                    Objects.requireNonNull(responseBody).byteStream(), typeRef);
+            log.info("Collected information about the partitions and hosts."
+                            + " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
+                    (int) partitionHostMappingResponse.values().stream().distinct().count());
```
torbsto, line 177: This is potentially somewhat costly. You could avoid unnecessary calculation by wrapping it in an if statement

common/src/main/java/com/bakdata/quick/common/api/client/StreamsStateHost.java
```diff
@@ -0,0 +1,72 @@
...
+        return String.format("http://%s/%s/%s",
+                this.host, this.config.getPath(), MirrorConfig.DEFAULT_PARTITION_INFO_PATH);
+    }
+
+    /**
+     * Provides a default strategy for calculating partitions.
+     *
+     * @return default partitioner
+     */
+    public static PartitionFinder getDefaultPartitionFinder() {
```
torbsto, line 69: Where do you use this? I think this should be handled by dependency injection

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,85 @@
...
+ * @param <K> the type of key
+ */
+public class PartitionRouter<K> implements Router<K> {
+
+    private final String topic;
+    private final Serde<K> keySerde;
+    private final PartitionFinder partitionFinder;
+    private final Map<Integer, String> partitionToHost;
+
+    private final Map<Integer, MirrorHost> partitionToMirrorHost = new HashMap<>();
```
torbsto, line 41: move into constructor

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,85 @@
...
+                           final PartitionFinder partitionFinder, final Map<Integer, String> partitionToHost) {
+        this.topic = topic;
+        this.keySerde = keySerde;
+        this.partitionFinder = partitionFinder;
+        this.partitionToHost = partitionToHost;
+        convertHostStringToMirrorHost();
+    }
+
+    private void convertHostStringToMirrorHost() {
+        for (final Map.Entry<Integer, String> entry : this.partitionToHost.entrySet()) {
```
torbsto, line 61: This comes down to personal preference, but I would use the streams API here and return a new Map

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionFinder.java
```diff
@@ -0,0 +1,32 @@
...
+public interface PartitionFinder {
+
+    /**
+     * Calculates the partition for a specific serialized key.
+     *
+     * @param serializedKey the byte representation of a key for which the partition is sought
+     * @param numPartitions the total number of partitions in a topic
+     * @return an int that represents a partition
+     */
+    int getForSerializedKey(final byte[] serializedKey, final int numPartitions);
```
torbsto, line 31: Could we remove `numPartitions` from the interface? It seems to me that it is an implementation detail of a particular `PartitionFinder` and shouldn't be part of the interface itself. Not too important though if you feel it isn't straightforward to implement

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,85 @@
...
+import com.bakdata.quick.common.api.model.mirror.MirrorHost;
+import com.bakdata.quick.common.config.MirrorConfig;
+import org.apache.kafka.common.serialization.Serde;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A router that leverages the fact that a mirror knows which partitions it stores
```
torbsto, line 28: To be precise, this leverages the fact that a mirror known which partitions which replica stores, not only itself

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,85 @@
...
+ * (it has a mapping between a partitions and a host),
```
torbsto, line 29:
```suggestion
 * (it has a mapping between partitions and a host),
```

common/src/main/java/com/bakdata/quick/common/util/KeySerdeValResolverWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.kafka.common.serialization.Serde;
+
+/**
+ * A wrapper for key serde and value type resolver.
+ * The wrapper is needed in order to return a single value from FetcherFactory.
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+@Getter
```
torbsto, line 30: You can use `@Value` here to save even more boilerplate

common/src/main/java/com/bakdata/quick/common/util/KeySerdeValResolverWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+public class KeySerdeValResolverWrapper<K, V> {
```
torbsto, line 31:
```suggestion
public class KeySerdeValueResolverWrapper<K, V> {
```

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/BaseMirrorClient.java
DawidNiezgodka: At first I did but then I thought "no, it doesn't seem right", so I check some classes you or Ramin wrote, and the first I found was QuickGraphQLInvocation which has import org.jetbrains.annotations.Nullable; so I thought this one is correct and edu.umd is wrong :P

common/src/main/java/com/bakdata/quick/common/api/client/StreamsStateHost.java
DawidNiezgodka: In `PartitionedMirrorClient` there are two constructors: One with PartitionFinder injected, another without. When no PartitionFinder is passed, then the constructor calls: `this.partitionFinder = StreamsStateHost.getDefaultPartitionFinder();`

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionFinder.java
DawidNiezgodka: Maybe we can make a separate issue out of it?

torbsto: No need, it's fine for now

common/src/main/java/com/bakdata/quick/common/api/client/MirrorClient.java
DawidNiezgodka: Actually there is no duplication: both BaseMirrorClient (BMC) and PartitionedMirrorClient (PMC) use MirrorRequestManager to manage their requests to the mirror. They use functions that are implemented there, i.e. delegate their request so that the logic does not have to be implemented in each of those classes. Previously we had duplication as a big part of the logic for making request was identical in BMC and PMC.

Regardless of this, BMC and DefaultMirrorClient (DMC) are almost identical, i.e. DMC uses BMC to delegate all its reqs so one could argue if the distinction between the two is reasonable now. However, should we change or extend DMC, the use of BMC might turn out fair-minded so I would leave it as it is.

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionFinder.java
DawidNiezgodka: Ok, so maybe an issue in quick-internal? Just to have in mind that this might be improved in the future.

common/src/main/java/com/bakdata/quick/common/api/client/MirrorClient.java
torbsto: Exactly, my question was about the relationship of BMC and DMC. In which scenario would the existence of BMC make sense? You've introduced it to avoid duplication in PMC and DMC, though this is now handled by the request manager.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorClient.java
```diff
@@ -20,18 +20,23 @@
...
 import com.bakdata.quick.common.resolver.TypeResolver;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
- * Default HTTP client for working with Quick mirrors.
+ * An abstract BaseMirrorClient that provides basic properties and functionalities for Mirror Clients.
```
torbsto, line 29: The old comment seems more fitting

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorClient.java
```diff
@@ -20,18 +20,23 @@
...
  *
  * @param <K> key type
  * @param <V> value type
  */
-
 public class DefaultMirrorClient<K, V> implements MirrorClient<K, V> {
 
-    private final MirrorClient<K, V> delegate;
+    protected final MirrorHost host;
```
torbsto, line 36: All variables can be private again

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorClient.java
```diff
@@ -65,64 +58,35 @@ public DefaultMirrorClient(final String topicName, final HttpClient client, fina
      * @param client       http client
      * @param typeResolver the value's {@link TypeResolver}
      */
-    public DefaultMirrorClient(final MirrorHost mirrorHost, final HttpClient client,
-                               final TypeResolver<V> typeResolver) {
+    public DefaultMirrorClient(final MirrorHost mirrorHost, final HttpClient client, final TypeResolver<V> typeResolver) {
         this.host = mirrorHost;
         this.client = client;
         this.parser = new MirrorValueParser<>(typeResolver, client.objectMapper());
+        this.mirrorRequestManager = new DefaultMirrorRequestManager(client);
```
torbsto, line 65: I'd prefer if this would also be injected

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -0,0 +1,84 @@
...
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+/**
+ * A default implementation of MirrorRequestManager.
+ */
+@Slf4j
+public class DefaultMirrorRequestManager implements MirrorRequestManager {
+
+    protected final HttpClient client;
```
torbsto, line 35:
```suggestion
    private final HttpClient client;
```

common/src/main/java/com/bakdata/quick/common/api/client/ParserFunction.java
```diff
@@ -25,7 +25,7 @@
  *
  * @param <T> type of the returned value
  */
-interface ParserFunction<T> {
+public interface ParserFunction<T> {
```
torbsto, line 28: Does this really have to be public?

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,177 @@
...
+        return keys.stream().map(this::fetchValue).collect(Collectors.toList());
+    }
+
+    @Override
+    public boolean exists(final K key) {
+        return this.fetchValue(key) != null;
+    }
+
+    private void initRouter() {
+        log.info("Initializing partition router...");
```
torbsto, line 152: add some contex: For which topic, which mirror?

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
DawidNiezgodka: Nope:
```
final String url = this.streamsStateHost.getPartitionToHostUrl();
        Map<Integer, String> partitionHostMappingResponse = new HashMap<>();
        try (final ResponseBody responseBody = requestManager.makeRequest(url)) {
            if (responseBody != null) {
                final TypeReference<Map<Integer, String>> typeRef = new TypeReference<>() {};
                partitionHostMappingResponse = this.client.objectMapper()
                    .readValue(Objects.requireNonNull(responseBody).byteStream(), typeRef);
                log.info("Collected information about the partitions and hosts."
                        + " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
                    (int) partitionHostMappingResponse.values().stream().distinct().count());
            } else {
                log.info("Response body was null. Returning an empty mapping.");
            }
            return partitionHostMappingResponse;
        } catch (final IOException e) {
            throw new InternalErrorException("There was a problem handling the response: " + e.getMessage());
        }
```

DawidNiezgodka: But probably the empty mapping is senseless, so I think I would rather throw an exception there.

DawidNiezgodka:
```
        try (final ResponseBody responseBody = requestManager.makeRequest(url)) {
            if (responseBody != null) {
                final TypeReference<Map<Integer, String>> typeRef = new TypeReference<>() {};
                final Map<Integer, String> partitionHostMappingResponse = this.client.objectMapper()
                    .readValue(Objects.requireNonNull(responseBody).byteStream(), typeRef);
                log.info("Collected information about the partitions and hosts."
                        + " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
                    (int) partitionHostMappingResponse.values().stream().distinct().count());
                return partitionHostMappingResponse;
            } else {
                throw new InternalErrorException("Response body was null. Returning an empty mapping.");
                
            }
```

torbsto: I was referring to ` (int) partitionHostMappingResponse.values().stream().distinct().count()` in the logging statement

DawidNiezgodka: Ok, I thought you were referring to the fact that makeRequest can return null - which I didn't cover btw.
Basically we can throw the info about the number of hosts away if you think it is costly. On the other hand, the number of replicas won't change dynamically (will it?), so I guess you won't compute this very often

torbsto: You can wrap it in  `if(log.isInfoEnabled) {...}`.

For the null case: Good catch, though `/partitions` should never return `null`. I'd also suggest exiting early, e.g.:
```java
if (responseBody == null) {
  throw new InternalErrorException(...)
}

// here conversion to map
```

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,177 @@
...
+    private final StreamsStateHost streamsStateHost;
+    private final HttpClient client;
+    private final Serde<K> keySerde;
+    private final String topicName;
+    private final PartitionFinder partitionFinder;
+    private final MirrorValueParser<V> parser;
+    private final MirrorRequestManager requestManager;
+
+    private Router<K> router;
+    private List<MirrorHost> knownHosts;
```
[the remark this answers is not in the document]
DawidNiezgodka: In order to avoid logic in constructor. The init of the router and the corresponding known hosts happens in the initRouter function where you assign specific values to these two fields.

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,177 @@
...
+    private void initRouter() {
+        log.info("Initializing partition router...");
+        final Map<Integer, String> response = makeRequestForPartitionHostMapping();
+        this.router = new PartitionRouter<>(this.keySerde, this.topicName, this.partitionFinder, response);
+        this.knownHosts = this.router.getAllHosts();
+    }
+
+    private Map<Integer, String> makeRequestForPartitionHostMapping() {
+        final String url = this.streamsStateHost.getPartitionToHostUrl();
+        Map<Integer, String> partitionHostMappingResponse = new HashMap<>();
```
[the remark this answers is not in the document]
DawidNiezgodka: With the early return (an exception) in case of failure we have only one place where it can be initialized so it can be final:

```
private Map<Integer, String> makeRequestForPartitionHostMapping() {
        final String url = this.streamsStateHost.getPartitionToHostUrl();
        try (final ResponseBody responseBody = requestManager.makeRequest(url)) {
            if (responseBody != null) {
                final Map<Integer, String> partitionHostMappingResponse = this.client.objectMapper()
                    .readValue(Objects.requireNonNull(responseBody).byteStream(), MAP_TYPE_REFERENCE);
                if (log.isInfoEnabled()) {
                    log.info("Collected information about the partitions and hosts."
                            + " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
                        (int) partitionHostMappingResponse.values().stream().distinct().count());
                }
                return partitionHostMappingResponse;
            } else {
                throw new InternalErrorException("Response body was null.");

            }
        } catch (final IOException e) {
            throw new InternalErrorException("There was a problem handling the response: " + e.getMessage());
        }
```

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -0,0 +1,84 @@
...
+    public DefaultMirrorRequestManager(final HttpClient client) {
+        this.client = client;
+    }
+
+    @Override
+    @Nullable
+    public <T> T sendRequest(final String url, final ParserFunction<T> parser) {
+        try {
+            final ResponseBody body = makeRequest(url);
+            final MirrorValue<T> mirrorValue;
```
torbsto, line 46: Why declare this variable here?

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Responsible for managing requests send to the Mirror.
+ */
+public interface MirrorRequestManager {
+
+    /**
+     * Responsible for making a request to a specific url and processing the result.
+     *
+     * @param url    a url for which a request is made
+     * @param parser parser
```
torbsto, line 31: please add a proper description

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param <T>    type
```
torbsto, line 32: see above; type of what?

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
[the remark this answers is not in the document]
torbsto: I don't think this is a good reason. You can either return those two values from the `initRouter` function or drop `initRouter`as it is only 4 lines.

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @param topicName     the name of the topic
+     * @param mirrorHost    host to use
+     * @param client        http client
+     * @param keySerde      the serde for the key
+     * @param valueResolver the value's {@link TypeResolver}
+     */
+    public PartitionedMirrorClient(final String topicName, final MirrorHost mirrorHost, final HttpClient client,
+                                   final Serde<K> keySerde, final TypeResolver<V> valueResolver) {
+
```
torbsto, line 88: You can also use  `this(...)` here

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,182 @@
...
+            this.streamsStateHost.getHost(), this.topicName);
+        final Map<Integer, String> response = makeRequestForPartitionHostMapping();
+        this.router = new PartitionRouter<>(this.keySerde, this.topicName, this.partitionFinder, response);
+        this.knownHosts = this.router.getAllHosts();
+    }
+
+    private Map<Integer, String> makeRequestForPartitionHostMapping() {
+        final String url = this.streamsStateHost.getPartitionToHostUrl();
+        try (final ResponseBody responseBody = requestManager.makeRequest(url)) {
+            if (responseBody != null) {
```
torbsto, line 165: I'd still prefer the early error as written before, see https://github.com/bakdata/quick/pull/38#discussion_r918844318

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param partitionFinder strategy for finding partitions
+     * @param partitionToHost partition to host mapping
+     */
+    public PartitionRouter(final Serde<K> keySerde, final String topic, final PartitionFinder partitionFinder,
+                           final Map<Integer, String> partitionToHost) {
+        this.topic = topic;
+        this.keySerde = keySerde;
+        this.partitionFinder = partitionFinder;
+        this.partitionToHost = partitionToHost;
+        this.partitionToMirrorHost = convertHostStringToMirrorHost();
```
torbsto, line 56:
```suggestion
        this.partitionToMirrorHost = this.convertHostStringToMirrorHost();
```

common/src/main/java/com/bakdata/quick/common/config/MirrorConfig.java
```diff
@@ -68,4 +70,10 @@ public MirrorConfig(final Optional<String> prefix, final Optional<String> path)
     public static MirrorConfig directAccess() {
         return new MirrorConfig(Optional.of(""), Optional.of(DEFAULT_MIRROR_HOST_PATH));
     }
+
+    public static MirrorConfig directAccessToStreamsState() {
+        return new MirrorConfig(Optional.of(""), Optional.of(DEFAULT_STREAMS_STATE_PATH));
+    }
+
```
torbsto, line 77: remove empty line

gateway/src/test/java/com/bakdata/quick/gateway/directives/rest/RestDirectiveWiringTest.java
```diff
@@ -53,9 +53,8 @@ class RestDirectiveWiringTest {
     void setUp() {
```
torbsto, line 53: Any reason you've reformatted this file? It's make reviewing pretty hard because of the noise

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListArgumentFetcherTest.java
```diff
@@ -41,6 +34,14 @@
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
+import java.util.ArrayList;
```
torbsto, line 37: wrong import order

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListFetcherTest.java
```diff
@@ -30,15 +28,18 @@
...
-import java.util.List;
-import java.util.Map;
 import lombok.Builder;
 import lombok.Data;
 import okhttp3.OkHttpClient;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import org.junit.jupiter.api.Test;
 
+import java.util.List;
```
torbsto, line 38: wrong import order

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/RestDataFetcherTest.java
```diff
@@ -31,15 +28,19 @@
...
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
 import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Test;
 
+import java.util.HashMap;
```
torbsto, line 37: see above :)

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,81 @@
...
+ * and thus can use this information to introduce routing based on the specific partition mapping.
+ *
+ * @param <K> the type of key
+ */
+public class PartitionRouter<K> implements Router<K> {
+
+    private final String topic;
+    private final Serde<K> keySerde;
+    private final PartitionFinder partitionFinder;
+    private final Map<Integer, String> partitionToHost;
```
torbsto, line 39: Do you need to keep this map around? Seems like you only need it for initializing

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,81 @@
...
+    private Map<Integer, MirrorHost> convertHostStringToMirrorHost() {
+        return this.partitionToHost.entrySet().stream().collect(
+            Collectors.toMap(Map.Entry::getKey, e -> new MirrorHost(e.getValue(), MirrorConfig.directAccess())));
+    }
+
+    @Override
+    public MirrorHost getHost(final K key) {
+        final byte[] serializedKey = this.keySerde.serializer().serialize(topic, key);
+        final int partition = partitionFinder.getForSerializedKey(serializedKey, this.partitionToMirrorHost.size());
+        if (partitionToMirrorHost.isEmpty() || !partitionToMirrorHost.containsKey(partition)) {
```
torbsto, line 68: Isn't `isEmpty()` covered by `containsKey(...)`? Whenever it's empty, `containsKey` will also return false

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,182 @@
...
+                final Map<Integer, String> partitionHostMappingResponse = this.client.objectMapper()
+                    .readValue(Objects.requireNonNull(responseBody).byteStream(), MAP_TYPE_REFERENCE);
```
torbsto, line 167:
```suggestion
                    .readValue(responseBody.byteStream(), MAP_TYPE_REFERENCE);
```
You've checked it above

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
DawidNiezgodka: There were 3 constructors and only one was used so I removed the unnecessary ones. One can add them later if they are needed

DawidNiezgodka: And concerning the initRouter(), I still have some doubts. On one hand, this doesnt look good to me:
    ```
public PartitionedMirrorClient(final String topicName, final MirrorHost mirrorHost, final HttpClient client,
                                   final Serde<K> keySerde, final TypeResolver<V> valueResolver,
                                   final PartitionFinder partitionFinder) {
        this.streamsStateHost = StreamsStateHost.fromMirrorHost(mirrorHost);
        this.client = client;
        this.keySerde = keySerde;
        this.topicName = topicName;
        this.partitionFinder = partitionFinder;
        this.parser = new MirrorValueParser<>(valueResolver, client.objectMapper());
        this.requestManager = new DefaultMirrorRequestManager(client);
        log.info("Initializing partition router for the mirror at: {} and the topic: {}.",
            this.streamsStateHost.getHost(), this.topicName);
        final Map<Integer, String> response = makeRequestForPartitionHostMapping();
        this.router = new PartitionRouter<>(this.keySerde, this.topicName, this.partitionFinder, response);
        this.knownHosts = this.router.getAllHosts();
    }
```
On the other hand, create an extra object and passing again all the needed info just to have these two fields seems like an overkill.
The problem with the logic in constructor is: if we want to overload the constructor later, we will have a code duplication. That was the reason to have it in a separate function.

File: gateway/src/test/java/com/bakdata/quick/gateway/directives/rest/RestDirectiveWiringTest.java (hunk 30 above)
Comment (DawidNiezgodka) [reply gateway/src/test/java/com/bakdata/quick/gateway/directives/rest/RestDirectiveWiringTest.java:53]: Hmm, I took a look at the file where I changed sth and than just applied checkstyle. Thought it is desired to do so.

Comment (torbsto) [reply gateway/src/test/java/com/bakdata/quick/gateway/directives/rest/RestDirectiveWiringTest.java:53]: It seems like you hit reformat twice, IntelliJ then reformats custom line breaks. Could you revert the changes?

File: common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java (hunk 26 above)
Comment (torbsto) [reply common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java:88]: A new class is ~6 lines, that's totally fine for this use case. 

Also regarding the logic in the constructor: That's why `this(...)` is preferred :)

File: common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java [hunk 37]
```diff
@@ -0,0 +1,79 @@
...
+
+    private Map<Integer, MirrorHost> convertHostStringToMirrorHost(final Map<Integer, String> partitionToHost) {
+        return partitionToHost.entrySet().stream().collect(
+            Collectors.toMap(Map.Entry::getKey, e -> new MirrorHost(e.getValue(), MirrorConfig.directAccess())));
+    }
+
+    @Override
+    public MirrorHost getHost(final K key) {
+        final byte[] serializedKey = this.keySerde.serializer().serialize(topic, key);
+        final int partition = partitionFinder.getForSerializedKey(serializedKey, this.partitionToMirrorHost.size());
```
torbsto, line 65: Couple of missing `this`

common/src/main/java/com/bakdata/quick/common/api/client/StreamsStateHost.java
```diff
@@ -0,0 +1,82 @@
...
+            MirrorConfig.DEFAULT_PARTITION_INFO_PATH);
+    }
+
+    /**
+     * Returns host-info.
+     *
+     * @return a string containing host info
+     */
+    public String getHost() {
+        return host;
```
torbsto, line 80: this

common/src/main/java/com/bakdata/quick/common/api/client/MirrorValueParser.java
```diff
@@ -33,20 +33,20 @@
  *
  * @param <V> type of the value
  */
-class MirrorValueParser<V> {
+public class MirrorValueParser<V> {
```
torbsto, line 36: I still don't see why this needs to be public now?

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListFetcherTest.java
torbsto: not resolved

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/RestDataFetcherTest.java
torbsto: not resolved

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -0,0 +1,79 @@
...
+        this.partitionToMirrorHost = this.convertHostStringToMirrorHost(partitionToHost);
+    }
+
+    private Map<Integer, MirrorHost> convertHostStringToMirrorHost(final Map<Integer, String> partitionToHost) {
+        return partitionToHost.entrySet().stream().collect(
+            Collectors.toMap(Map.Entry::getKey, e -> new MirrorHost(e.getValue(), MirrorConfig.directAccess())));
+    }
+
+    @Override
+    public MirrorHost getHost(final K key) {
```
torbsto, line 63: Maybe `findHost` or `calcualteHost`? I think it can be misleading to use `get...` for a non-trivial task

gateway/src/test/java/com/bakdata/quick/gateway/directives/rest/RestDirectiveWiringTest.java
DawidNiezgodka: Ok, I'll just copy it from master.

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListArgumentFetcherTest.java
DawidNiezgodka: I will copy from master

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListFetcherTest.java
DawidNiezgodka: same as above

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -0,0 +1,81 @@
...
+            throw new MirrorException("Not able to parse content", HttpStatus.INTERNAL_SERVER_ERROR, exception);
+        }
+    }
+
+    @Override
+    @Nullable
+    public ResponseBody makeRequest(final String url) {
+        final Request request = new Request.Builder().url(url).get().build();
+        try {
+            final Response response = this.client.newCall(request).execute();
```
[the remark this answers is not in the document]
DawidNiezgodka: When you do this, you will explicitly close the input stream, and you won't be able to read from it in the PartitionedMirrorClient.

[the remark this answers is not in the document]
torbsto: Ah true, could you add a comment for this?

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,135 @@
...
+        try (final ResponseBody responseBody = this.requestManager.makeRequest(url)) {
+            if (responseBody == null) {
+                throw new InternalErrorException("Response body was null.");
+            }
+            final Map<Integer, String> partitionHostMappingResponse = this.client.objectMapper()
+                .readValue(responseBody.byteStream(), MAP_TYPE_REFERENCE);
+            this.logPartitionHostInfoIfLoggingEnabled(partitionHostMappingResponse);
+            return partitionHostMappingResponse;
+        } catch (final IOException e) {
+            throw new InternalErrorException("There was a problem handling the response: " + e.getMessage());
```
raminqaf, line 124:
```suggestion
        } catch (final IOException exception) {
            throw new MirrorException("There was a problem handling the response: ", HttpStatus.INTERNAL_SERVER_ERROR, exception);
```

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public boolean exists(final K key) {
+        return this.fetchValue(key) != null;
+    }
+
+    private Map<Integer, String> makeRequestForPartitionHostMapping() {
+        final String url = this.streamsStateHost.getPartitionToHostUrl();
+        try (final ResponseBody responseBody = this.requestManager.makeRequest(url)) {
+            if (responseBody == null) {
+                throw new InternalErrorException("Response body was null.");
```
raminqaf, line 117: This is still a MirrorClient. I would do it like the DefaultMirrorClient:
```suggestion
                throw new MirrorException("Response body was null.", HttpStatus.INTERNAL_SERVER_ERROR);
```

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    private void logPartitionHostInfoIfLoggingEnabled(Map<Integer, String> partitionHostMappingResponse) {
+        if (log.isInfoEnabled()) {
+            log.info("Collected information about the partitions and hosts." +
+                    " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
+                (int) partitionHostMappingResponse.values().stream().distinct().count());
+        }
+    }
```
raminqaf, line 134: See above

common/src/main/java/com/bakdata/quick/common/api/client/StreamsStateHost.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package com.bakdata.quick.common.api.client;
+
+import com.bakdata.quick.common.api.model.mirror.MirrorHost;
+import com.bakdata.quick.common.config.MirrorConfig;
+
+/**
+ * Provides information about the host that enables access to the endpoint
+ * that delivers info about kafka streams app state.
```
raminqaf, line 24:
```suggestion
 * that delivers info about Kafka Streams app state.
```

common/src/main/java/com/bakdata/quick/common/config/MirrorConfig.java
```diff
@@ -68,4 +70,8 @@ public MirrorConfig(final Optional<String> prefix, final Optional<String> path)
     public static MirrorConfig directAccess() {
         return new MirrorConfig(Optional.of(""), Optional.of(DEFAULT_MIRROR_HOST_PATH));
     }
+
+    public static MirrorConfig directAccessToStreamsState() {
```
raminqaf, line 74: Javadocs?

common/src/test/java/com/bakdata/quick/common/api/client/routing/PartitionRouterTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Test for Partition Router.
+ */
+@MicronautTest
+public class PartitionRouterTest {
+
+    @Test
+    void shouldBeInitializedWithTwoDifferentHosts() {
+
```
raminqaf, line 35: remove line

common/src/test/java/com/bakdata/quick/common/api/client/routing/PartitionRouterTest.java
```diff
@@ -0,0 +1,50 @@
...
+        final Map<Integer, String> elements = Map.of(1, "host1", 2, "host2");
+        final Router<String> partitionRouter =
+            new PartitionRouter<>(Serdes.String(), "dummy", TestUtils.getMockPartitionFinder(), elements);
+        assertThat(partitionRouter.getAllHosts()).hasSize(2);
+    }
+
+    @Test
+    void shouldReturnCorrectHostForGivenPartition() {
+
```
raminqaf, line 44: remove line

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java
```diff
@@ -0,0 +1,46 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import okhttp3.ResponseBody;
+
+/**
+ * Responsible for managing requests send to the Mirror.
+ */
+public interface MirrorRequestManager {
+
+    /**
+     * Responsible for making a request to a specific url and processing the result.
```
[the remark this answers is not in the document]
DawidNiezgodka: Yea, I also didn't like that. That's why it has been changed. The results are in another branch: https://github.com/bakdata/quick/blob/8ed54d4c504a553b6d1c0e841d8651261f9de43e/common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return the value from a mirror value wrapper
+     */
+    @Nullable
+    <T> T sendRequest(final String url, final ParserFunction<T> parser);
+
+    /**
+     * Submits a request and processes the response. Throws an exception in case of various errors.
+     *
+     * @param url a url for which a request is made
+     * @return response body if successful; null if resource has not been found
```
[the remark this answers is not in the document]
DawidNiezgodka: @torbsto - I asked you the same question, and you said that sometimes it is ok to return null (see Slack, 6.07)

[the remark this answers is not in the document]
torbsto: Yes, in case the GraphQL query defines the return type as nullable we have to propagate this properly and cannot throw an exception

torbsto (approved):

raminqaf (approved):

PR #42, opened 2022-07-06
Remove required avro namespace from Helm chart

torbsto:
This removes "avro.namespace" from the Helm chart, as it is no longer required

### DIFF
--- a/docs/docs/user/reference/helm-chart.md
+++ b/docs/docs/user/reference/helm-chart.md
@@ -1,6 +1,7 @@
 # Quick Helm chart
 
 Below you can find the default `value.yaml` of Quick's Helm chart.
+
 ```yaml
 image:
   # The base repository of the images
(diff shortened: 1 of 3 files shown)


docs/docs/user/reference/helm-chart.md
```diff
@@ -60,4 +55,7 @@ quickConfig:
   QUICK_TOPIC_REGISTRY_PARTITIONS: "3"
   QUICK_TOPIC_REGISTRY_REPLICATION_FACTOR: "1"
   QUICK_SCHEMA_FORMAT: "Avro"
+  # The namespace used for created avro namespaces
+  # see https://avro.apache.org/docs/current/spec.html
+  QUICK_SCHEMA_AVRO_FORMAT: ""
```
raminqaf, line 60: I guess this should be `QUICK_SCHEMA_AVRO_NAMESPACE`

raminqaf: Do we want to add a comment for Protobuf here? Like this:
```yaml
# For Protobuf
# QUICK_SCHEMA_FORMAT: "Protobuf"
# QUICK_SCHEMA_PROTOBUF_PACKAGE: ""
 ```

raminqaf (approved):

docs/docs/user/reference/helm-chart.md
torbsto: Maybe we should just link https://bakdata.github.io/quick/0.6/user/reference/configuration/ and remove all examples?

PR #44, opened 2022-07-12
Add release docs for 0.7

torbsto:


docs/docs/roadmap.md
```diff
@@ -6,22 +6,17 @@ and [project boards](https://github.com/bakdata/quick/projects).
...
 Development: Q3 2022
 
 * Improved gateway performance: Pre-computation of key's location
-* Kafka-Streams library for custom REST APIs in Quick
+* Kafka 3.0 support
 
 ### Further ideas
 
+* Kafka-Streams library for custom REST APIs in Quick
+* More complex query arguments, like ranges and filter
```
DawidNiezgodka, line 19: Only a single filter? Or ranges and filter**s**?

docs/docs/roadmap.md
```diff
@@ -6,22 +6,17 @@ and [project boards](https://github.com/bakdata/quick/projects).
...
-Development: Q2 2022
-
-* Protobuf support
-* Kafka 3.0 support
-
 ### 0.8
 
 Development: Q3 2022
 
 * Improved gateway performance: Pre-computation of key's location
```
DawidNiezgodka, line 13: This was already here but shouldn't it be either of a key's location or keys' location depending on whether singular or plural is meant.

docs/docs/roadmap.md
torbsto: - ranges -> two arguments, e.g. from, to
- filter -> one argument, e.g. smaller than

DawidNiezgodka: That's really a trifle - if filter is countable and we have a single one, then we might write "a filter" and not just filter.

torbsto: filters it is then :smile:

PR #46, opened 2022-07-15
A mechanism for automatic update of partition mapping info

### LINKED ISSUE #39: Add a mechanism for the dynamic update of partition-host mapping to the PartitionRouter (PartitionedMirrorClient)
Currently we fetch information about the partition-host mapping once during the initialisation of the PartitionRouter. Should the mapping change, for example because of the addition of another replica, some cache-misses will occur as there is no update-mechanism. Thus, it is advisable to introduce such a mechanism, i.e. the possibility to update the mapping dynamically. 

A possible approach:
1. If there is a host-miss, the request is forwarded to the correct replica without any error,
2. In addition, a prompt to update the cache is written into a header,
3. The gateway uses this info to forward it to the router which reinits (updates) the mapping.

DawidNiezgodka:
This PR resolves #39.


mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -126,35 +125,40 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
+            new DefaultMirrorClient<>(mirrorHost, this.client, this.valueResolver,
+                new DefaultMirrorRequestManager(this.client));
         // TODO: don't bother deserializing
         final V value = mirrorClient.fetchValue(key);
 
         if (value == null) {
             throw new NotFoundException("Key not found");
         }
 
+        this.client.okHttpClient().interceptors().add(chain -> {
```
torbsto, line 152: This will tell the other mirror replica that there was a cache miss, but we want to inform the caller of this mirror about it. So you have to manipulate the response, not the request. See https://docs.micronaut.io/2.5.4/guide/#requestResponse

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -121,27 +136,30 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
     @Override
-    public Single<MirrorValue<List<V>>> getAll() {
+    public Single<HttpResponse<MirrorValue<List<V>>>> getAll() {
         // For now, we only consider the local state!
         final ReadOnlyKeyValueStore<K, V> store = this.streams.store(this.storeQueryParameters);
-        return Flowable.fromIterable(store::all)
-            .map(keyValue -> keyValue.value)
-            .toList()
-            .map(MirrorValue::new);
+        List<V> values = new ArrayList<>();
```
torbsto, line 159: The previous way is definitely preferable because it uses the reactive runtime underneath and doesn't block on IO. I think you could just modify the final call to `map` to create a `MirrorValue` wrapped in a `HttpResponse`

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -61,6 +63,8 @@
     private final TypeResolver<K> keyResolver;
     private final StoreQueryParameters<ReadOnlyKeyValueStore<K, V>> storeQueryParameters;
 
+    private boolean updateCache = false;
```
torbsto, line 66: This doesn't work if because it's not thread-safe.

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -78,14 +82,26 @@ public KafkaQueryService(final QueryContextProvider contextProvider, final Quick
...
+                QuickConstants.getUpdateMappingHeader(), QuickConstants.getCacheUpdateMessage()
+            );
+            updateCache = false;
+        } else {
+            response = HttpResponse.created(new MirrorValue<>(value));
+        }
+        return Single.just(response);
+    }
+
+    private V internalGet(final String rawKey) {
```
torbsto, line 104: I don't see the advantage of using an `internalGet`, especially with a mutable member variable.
If you want to use the `internalGet`, why not just return the final `HttpReponse<...>`?

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -121,27 +136,30 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
     @Override
-    public Single<MirrorValue<List<V>>> getValues(final List<String> keys) {
-        return Flowable.fromIterable(keys)
-            .flatMapSingle(this::get)
-            .map(MirrorValue::getValue)
-            .toList()
-            .map(MirrorValue::new);
+    public Single<HttpResponse<MirrorValue<List<V>>>> getValues(final List<String> keys) {
+
+        MirrorValue<List<V>> mirrorValue =
```
torbsto, line 145: Same as above: Try to use the reactive API

DawidNiezgodka:
> > In order to account for the fact that we might have a header that encourages us to update the routing info, I introduced a ResponseWrapper that consists of a nullable ResponseBody and Optional that can be filled with if a header is available,
> 
> Have you thought about using the `Response` itself? It seems like the easier choice to me. Am I missing something?

Ye, I was thinking about it and even tried it out but somehow the dealing with nulls seemed more complicated than when you have it in a wrapper class. I can go back to this idea though if it is desired :)

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
DawidNiezgodka: I introduced it because `getValues` calls `get` multiple times, and I thought that it would be easier to have a separate function that return just V and not `Single<Http<MirrorValue<V>>>>`. Hmm but yes, we can have multiple HttpResponses from each get, extract all the values, and check if any of Responses had the update-cache-header set.

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
DawidNiezgodka: Ye, its time to change the way of thinking from a single threaded apps to multi threaded 👯‍♂️

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -124,11 +132,17 @@ private Map<Integer, String> makeRequestForPartitionHostMapping() {
...
-            log.info("Collected information about the partitions and hosts." +
-                    " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
+            log.info("Collected information about the partitions and hosts."
+                    + " There are {} partitions and {} distinct hosts", partitionHostMappingResponse.size(),
                 (int) partitionHostMappingResponse.values().stream().distinct().count());
         }
     }
+
+    private void updateRouterInfo() {
+        log.info("Updating router information");
```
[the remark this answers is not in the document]
DawidNiezgodka: This was already changed in another branch concerning Router but they have not been rebased yet.

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
DawidNiezgodka: This one is tough .

common/src/main/java/com/bakdata/quick/common/util/QuickConstants.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A utility class that contains various constants used in Quick.
+ */
+public class QuickConstants {
+
+    private QuickConstants() {}
+
+    private static final String UPDATE_PARTITION_HOST_MAPPING_HEADER = "X-Cache-Update";
+    private static final String CACHE_UPDATE_MESSAGE = "There was a cache miss. Please update routing information.";
```
torbsto, line 27: I'd argue that we don't need such a message because this is a technical implementation. I suggest completely removing the message: If the header exists, there was a cache-miss

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        this.server.enqueue(new MockResponse().setBody(body)
+            .setHeader(QuickConstants.getUpdateMappingHeader(), "There was a cache miss. Please update mapping"));
+
+        final String routerBody = TestUtils.generateBodyForRouterWith(Map.of(1, host, 2, host, 3, host));
+        this.server.enqueue(new MockResponse().setBody(routerBody));
+        final TopicData topic = this.topicDataClient.fetchValue(DEFAULT_TOPIC);
+        assertThat(Objects.requireNonNull(topic).getName()).isEqualTo(DEFAULT_TOPIC);
+
+
```
torbsto, line 82: remove

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void initRouterAndMirror() throws JsonProcessingException {
+        final String routerBody = TestUtils.generateBodyForRouterWith(Map.of(1, host, 2, host));
+        // first request sets the mapping with partitions 1 and 2
+        this.server.enqueue(new MockResponse().setBody(routerBody));
+        this.topicDataClient = new PartitionedMirrorClient<>(DEFAULT_TOPIC, mirrorHost, client, Serdes.String(),
+            new KnownTypeResolver<>(TopicData.class, this.mapper), TestUtils.getPartitionFinderForMappingUpdateTest());
+    }
+
+    @Test
+    void shouldReadHeaderAddUpdateRouter() throws JsonProcessingException {
```
torbsto, line 68:
```suggestion
    void shouldReadHeaderAndUpdateRouter() throws JsonProcessingException {
```

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,85 @@
...
+        // second request is the call to the MirrorClients fetchValue but as we set a header
+        // in the response this should trigger the third call which updates mapping of the router
+
+        this.server.enqueue(new MockResponse().setBody(body)
+            .setHeader(QuickConstants.getUpdateMappingHeader(), "There was a cache miss. Please update mapping"));
+
+        final String routerBody = TestUtils.generateBodyForRouterWith(Map.of(1, host, 2, host, 3, host));
+        this.server.enqueue(new MockResponse().setBody(routerBody));
+        final TopicData topic = this.topicDataClient.fetchValue(DEFAULT_TOPIC);
+        assertThat(Objects.requireNonNull(topic).getName()).isEqualTo(DEFAULT_TOPIC);
```
torbsto, line 80: I don't see how this tests if the underlying mapping was actually updated?

mirror/src/main/java/com/bakdata/quick/mirror/MirrorController.java
```diff
@@ -47,7 +48,7 @@ public MirrorController(final QueryService<V> queryService) {
      * Fetches values for the given id.
      */
     @Get("/{key}")
-    public Single<MirrorValue<V>> get(@PathVariable("key") final String keyString) {
+    public Single<? extends HttpResponse<MirrorValue<V>>> get(@PathVariable("key") final String keyString) {
```
torbsto, line 51: Any reason you're using a wildcard here but not for other routes?

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -105,9 +108,9 @@ public Single<MirrorValue<V>> get(final String rawKey) {
         // forward request if a different application is responsible for the rawKey
         if (!metadata.activeHost().equals(this.hostInfo) && !metadata.standbyHosts().contains(this.hostInfo)) {
             log.info("Forward request to {}", metadata.activeHost());
-            return Single.fromCallable(() -> this.fetch(metadata.activeHost(), key))
-                .map(MirrorValue::new)
-                .subscribeOn(Schedulers.io());
+            // The cast in the second map function is possible because MutableHttpResponse is of the same type
```
torbsto, line 111: This comment seems outdated. There is no `map` nor `MutableHttpResponse`

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +125,40 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
     @Override
-    public Single<MirrorValue<List<V>>> getValues(final List<String> keys) {
-        return Flowable.fromIterable(keys)
+    public Single<HttpResponse<MirrorValue<List<V>>>> getValues(final List<String> keys) {
+        return Observable.fromIterable(keys)
             .flatMapSingle(this::get)
-            .map(MirrorValue::getValue)
             .toList()
-            .map(MirrorValue::new);
+            .map(listOfResponses -> {
```
torbsto, line 136: Extract the lambda into its own method please

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +125,40 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
     @Override
-    public Single<MirrorValue<List<V>>> getAll() {
+    public Single<HttpResponse<MirrorValue<List<V>>>> getAll() {
         // For now, we only consider the local state!
         final ReadOnlyKeyValueStore<K, V> store = this.streams.store(this.storeQueryParameters);
         return Flowable.fromIterable(store::all)
             .map(keyValue -> keyValue.value)
             .toList()
-            .map(MirrorValue::new);
+            .map(e -> HttpResponse.created(new MirrorValue<>(e)));
```
torbsto, line 157: Please rename `e` into something less abstract

mirror/src/test/java/com/bakdata/quick/mirror/MirrorApplicationIntegrationTest.java
```diff
@@ -89,7 +89,7 @@ void test() throws InterruptedException {
                 .atMost(Duration.ofSeconds(12))
                 .untilAsserted(() -> when().get("http://" + this.hostConfig.toConnectionString() + "/mirror/{id}", "key1")
                 .then()
-                .statusCode(200)
+                .statusCode(201)
```
torbsto, line 92: Why did this change? 201 seems wrong to me

common/src/main/java/com/bakdata/quick/common/api/client/routing/Router.java
```diff
@@ -40,4 +41,6 @@
      * @return a list of hosts
      */
     List<MirrorHost> getAllHosts();
+
+    void update(Map<Integer, String> updatedRoutingInfo);
```
torbsto, line 45:
```suggestion
    void updateRountingInfo(Map<Integer, String> updatedRoutingInfo);
```

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -136,4 +144,10 @@ private Map<Integer, String> makeRequestForPartitionHostMapping() {
                 HttpStatus.INTERNAL_SERVER_ERROR, exception);
         }
     }
+
+    private void updateRouterInfo() {
+        log.info("Updating router information");
```
torbsto, line 149: Please add context

torbsto: I also think this should be a debug level

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Optional;
+import lombok.Getter;
+import lombok.Setter;
+import okhttp3.ResponseBody;
+
+/**
+ * A wrapper for a response from a call to made with the HttpClient.
+ * It consists of a response body extracted from the response and an optional field
```
torbsto, line 28:
```suggestion
 * A wrapper for a response from a call to made with the HttpClient.
 
 * <p>
 * It consists of a response body extracted from the response and an optional field
```

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * for keeping the information about the custom X-Cache-Update header. This header is used to
+ * signal the need for an update of the mapping between partitions and mirror hosts.
+ */
+@Getter
+@Setter
```
torbsto, line 33: Please use `@Value`

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+public class ResponseWrapper {
+
+    @Nullable
+    private ResponseBody responseBody;
+    @NonNull
```
torbsto, line 38: Not needed

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -38,22 +40,8 @@ public DefaultMirrorRequestManager(final HttpClient client) {
...
-        } catch (final IOException exception) {
-            throw new MirrorException("Not able to parse content", HttpStatus.INTERNAL_SERVER_ERROR, exception);
-        }
-    }
-
-    @Override
-    @Nullable
-    public ResponseBody makeRequest(final String url) {
+    public ResponseWrapper makeRequest(final String url) {
+        final ResponseWrapper responseWrapper = new ResponseWrapper();
```
torbsto, line 44: As I've noted above, please make this immutable

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -65,9 +53,9 @@ public ResponseBody makeRequest(final String url) {
         try {
             final Response response = this.client.newCall(request).execute();
             if (response.code() == HttpStatus.NOT_FOUND.getCode()) {
-                return null;
+                responseWrapper.setResponseBody(null);
```
torbsto, line 56: This could still be a cache miss

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/util/QuickConstants.java
DawidNiezgodka: But isn't a header a key-value pair? 
The `CACHE_UPDATE_MESSAGE` was thought of as a value thereof.
For example, if you have Keep-Alive header, you have `Keep-Alive: 100`. In a similar fashion, I thought that we would have a (custom) header, `X-Cashe-Update: There was a cache miss`.

DawidNiezgodka: The API itself forces you to put a key and a value there, i.e.: `header(CharSequence name, CharSequence value)`. What would you like to put as a value parameter?

torbsto: This [StackOverflow question](https://stackoverflow.com/questions/58854683/whats-the-standard-format-for-representing-true-and-false-values-in-an-http-hea) gives a couple of possible options, I'm fine with all of them. My point is just that `There was a cache miss. Please update routing information.` is like a user facing description and isn't needed here

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
DawidNiezgodka: 1. You start with the following mapping: 1->host, 2->host. This is the mapping the router is initialised with. It is fetched by the first `server.enqueue` in `@BeforeEach`.
2. When you make a call to `this.topicDataClient.fetchValue(),` you'll make a request in `DefaultMirrorRequestManager`. Since we create a mocked response with header, the control statement at the line 71 will be called and a header will be set in the `ResponseWrapper`. Because of this `this.updateRouterInfo();` at 90 of `PartitionedMirrorClient` will be called. Eventually, you will get a new mapping: 1->host, 2->host, 3->host. 
3. Now, in the first call to `this.topicDataClient.fetchValue`, the returned partition is 2. If we made a consecutive call to to `this.topicDataClient.fetchValue`, the returned partition will be 3. Why do we get 2 and then 3? This is a way of functioning of the custom `PartitioningRouter` that was made for the purpose of the test, see `PartitionFinderForUpdateMappingTest`,
 If the info had not been updated, you would have received `IllegalStateException`.
4. But ye, the test is incomplete, because we have to add another `this.topicDataClient.fetchValue` in order to be sure that the exception is not thrown. If you add it, it works.

DawidNiezgodka: I will add this description to the test, if you say it's ok

torbsto: I'd change [PartitionFinderForUpdateMappingTest.java](https://github.com/bakdata/quick/pull/46/files/adb9a65cd1e9215b57602dfe8f42e7c84fcad938#diff-c6c4046ffa9bbf20608e4e24453ec8048df27cec992e6ad8a0374ea4852c05ff) to be explicit. Maybe something similar to the `enque` of the MockWebServer for the partition to be returned next.

mirror/src/test/java/com/bakdata/quick/mirror/MirrorApplicationIntegrationTest.java
DawidNiezgodka: It no longer works with 200 after we have introduced HttpResponse as a return value. To be honest, I dunno why, but maybe it treats it different that just returning a value.

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
DawidNiezgodka: Ok, I changed it but I'm not sure if this is what you meant.

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+ * for keeping the information about the custom X-Cache-Update header. This header is used to
+ * signal the need for an update of the mapping between partitions and mirror hosts.
+ * </p>
+ */
+@Value
+public class ResponseWrapper {
+
+    @Nullable
+    ResponseBody responseBody;
+    @NonNull
```
torbsto, line 39: Not resolved, see https://github.com/bakdata/quick/pull/46/files#r934171886

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+    Optional<String> updateCacheHeader;
```
torbsto, line 40: What's the benefit of using `Optional<String>` here? Are we ever interested in the value inside the optional? I'd have expected a boolean instead

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
torbsto: Let's talk about it synchronously, I had something different in mind :smile:

common/src/test/java/com/bakdata/quick/common/testutils/PartitionFinderForUpdateMappingTest.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.bakdata.quick.common.testutils;
+
+import com.bakdata.quick.common.api.client.routing.PartitionFinder;
+
+/**
+ * A custom PartitionFinder for testing the update mechanism of the PartitionRouter.
+ */
+public class PartitionFinderForUpdateMappingTest implements PartitionFinder {
```
torbsto, line 24:
```suggestion
public class TestPartitionFinder implements PartitionFinder {
```

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +124,56 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
-            .map(MirrorValue::new);
+            .map(this::transformValuesAndCreateHttpResponse);
+    }
+
+    /**
+     * Transforms a list of HttpResponses of MirrorValue of a specific type into
+     * a single HttpResponse of MirrorValue with a list of values of that type.
+     * Furthermore, if a header is present in one of the HttpResponses (function argument), a HTTP Header
+     * that informs about the Cache-Miss is set. Because of this possibility, the function returns MutableHttpResponse
+     * and not just HttpResponse. However, MutableHttpResponse is of type HttpResponse so there is no clash
```
torbsto, line 143: > However, MutableHttpResponse is of type HttpResponse so there is no clash with the QueryService interface that uses HttpResponse.

No need to document this, it's pretty intuitive that this is the case

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +124,56 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
+        final boolean headerSet = listOfResponses.stream()
+            .anyMatch(response -> response.header(HeaderConstants.getCacheMissHeaderName()) != null);
+        final List<V> values = listOfResponses.stream()
+            .map(response -> Objects.requireNonNull(response.body()).getValue()).collect(Collectors.toList());
+        if (headerSet) {
+            return HttpResponse.created(new MirrorValue<>(values))
+                .header(HeaderConstants.getCacheMissHeaderName(), HeaderConstants.getCacheMissHeaderValue())
+                .status(200);
+        } else {
+            return HttpResponse.created(new MirrorValue<>(values)).status(200);
```
torbsto, line 161: You can extract this as a variable and reuse it above

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +124,56 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
+     * @param listOfResponses a list of HttpResponses obtained from multiple calls to get(key),
+     *                        see getValues for the details
+     * @return MutableHttpResponse, possibly with a Cache-Miss Header set
+     */
+    private MutableHttpResponse<MirrorValue<List<V>>> transformValuesAndCreateHttpResponse(
+        final List<HttpResponse<MirrorValue<V>>> listOfResponses) {
+        final boolean headerSet = listOfResponses.stream()
+            .anyMatch(response -> response.header(HeaderConstants.getCacheMissHeaderName()) != null);
+        final List<V> values = listOfResponses.stream()
+            .map(response -> Objects.requireNonNull(response.body()).getValue()).collect(Collectors.toList());
```
torbsto, line 155:
```suggestion
            .map(response -> Objects.requireNonNull(response.body()).getValue())
            .collect(Collectors.toList());
```

common/src/test/java/com/bakdata/quick/common/testutils/TestUtils.java
```diff
@@ -55,4 +55,8 @@ public static String generateBodyForRouterWith(final Map<Integer, String> elemen
     public static PartitionFinder getMockPartitionFinder() {
         return (serializedKey, numPartitions) -> 1;
     }
+
+    public static PartitionFinder getPartitionFinderForMappingUpdateTest() {
```
torbsto, line 59: I don't really see the benefit of this method?

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -84,18 +86,25 @@ public PartitionedMirrorClient(final String topicName, final MirrorHost mirrorHo
...
     @Override
     public List<V> fetchAll() {
         final List<V> valuesFromAllHosts = new ArrayList<>();
         for (final MirrorHost host : this.knownHosts) {
-            final List<V> valuesFromSingleHost =
-                Objects.requireNonNullElse(this.requestManager.sendRequest(host.forAll(), this.parser::deserializeList),
-                    Collections.emptyList());
-            valuesFromAllHosts.addAll(valuesFromSingleHost);
+            final ResponseWrapper response = this.requestManager.makeRequest(host.forAll());
+            if (response != null) {
```
torbsto, line 102: Can the response even be null now that you use a wrapper?

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -79,9 +63,37 @@ public ResponseBody makeRequest(final String url) {
...
+                return parser.parse(responseWrapper.getResponseBody().byteStream()).getValue();
+            }
+            return null;
+        } catch (final IOException exception) {
+            throw new MirrorException("Not able to parse content", HttpStatus.INTERNAL_SERVER_ERROR, exception);
+        }
+    }
+
+    /**
+     * Checks if the X-Cache-Update header has been set.
```
raminqaf, line 89: Could you please add a `@link` tag and point out the X-Cache-Update to the `HeaderConstants` class?

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * 1) We start with the following mapping: 1->host, 2->host. This is the mapping the router is initialised with.
+     * It is fetched by the first server.enqueue in @BeforeEach.
+     * 2) When we make a call to this.topicDataClient.fetchValue(), we'll make a request in
+     * DefaultMirrorRequestManager (DMRM).
+     * Since we create a mocked response with a header, the control statement at the line 71 of DMRM will be called
+     * and the header will be set in the ResponseWrapper.
+     * 3) Because of this, this.updateRouterInfo(); at 90 of PartitionedMirrorClient will be called.
```
raminqaf, line 75: I would remove the line numbers in the comments

raminqaf (commented):
looks good to me just some small adjustments

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+ * A wrapper for a response from a call to made with the HttpClient.
+ *
+ * <p>
+ * It consists of a response body extracted from the response and a boolean that indicates
+ * whether the X-Cache-Update header has been set. This header is used to
+ * signal the need for an update of the mapping between partitions and mirror hosts.
+ * </p>
+ */
+@Value
+public class ResponseWrapper {
```
torbsto, line 33: I'd create a static creator `fromResponse(Response response)` that includes the corresponding logic for setting the header flag

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,103 @@
...
+    void initRouterAndMirror() throws JsonProcessingException {
+        final String routerBody = TestUtils.generateBodyForRouterWith(Map.of(1, host, 2, host));
+        // first request sets the mapping with partitions 1 and 2
+        this.server.enqueue(new MockResponse().setBody(routerBody));
+        this.topicDataClient = new PartitionedMirrorClient<>(DEFAULT_TOPIC, mirrorHost, client, Serdes.String(),
+            new KnownTypeResolver<>(TopicData.class, this.mapper), new TestPartitionFinder(List.of(2, 3)));
+    }
+
+    /**
+     * 1) We start with the following mapping: 1->host, 2->host. This is the mapping the router is initialized with.
```
torbsto, line 69: Generally, if you need such a long comment to explain a basic test, I'd suggest rather trying to rewrite the test to be more expressive. 
In this case, the problem is just that the relationship between the first router mapping, and the partition of the second call is hard to find. Therefore, my previous suggestion was to add a method `enque` for the `TestPartitionFinder`. It lets you start with `enque(2)` to signal that the first response is routed to the host of partition 2. Then, prior to the second request, you could call `enque(3)` to signal that the next partition returned is 3. 
It would maybe also be helpful to first test that if it's not updated, an error is being thrown.

torbsto (commented):
This handles the case when a new replica is added. What about a replica being removed?

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
DawidNiezgodka: Ok, but I'm not 100% convinced that it helps to understand the test. Someone might think: _"Ok, there's a partition finder, there is some queue, they enqueue 2, and then 3, but wait, this was supposed to be the test of router not partition finder, etc."._
I think that someone who reads the test must dig deeper to connect the dots. I agree that the comment is wordy, but I intended to help a potential reader find appropriate fragments in the code to facilitate understanding.
I guess that the underlying problem is that the router's logic is concealed. We don't interact with it directly. Instead, we influence its behavior via playing with PartitionFinder. Generally, it is difficult to grasp what happens behind the scenes without analyzing the code in detail.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +35,61 @@
...
         this.client = client;
+        this.fallbackServiceHost = fallbackServiceHost.plainUrl();
     }
 
     @Override
-    @Nullable
-    public <T> T sendRequest(final String url, final ParserFunction<T> parser) {
+    public ResponseWrapper makeRequest(final String url) {
+        final Request request = new Request.Builder().url(url).get().build();
+        /*
```
torbsto, line 48: Please us always line comments except for JavaDoc

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +35,61 @@
...
         try {
-            final ResponseBody body = this.makeRequest(url);
-            if (body != null) {
-                return parser.parse(body.byteStream()).getValue();
-            }
-            return null;
+            final Response response = this.client.newCall(request).execute();
+            return ResponseWrapper.fromResponse(response);
         } catch (final IOException exception) {
+            /*
```
torbsto, line 58: Extract all of this into a method

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +35,61 @@
...
+             * This code covers a situation where a replica is removed, and we can no longer
+             * reach the host for a given partition. The k8s-service (Load Balancer) is used
+             * as a fallback in such a case. It might also happen that the Load Balancer itself is not reachable.
+             * If this happens, we throw an exception. We also set the X-Cache-Update header to a response
+             * to immediately update the mapping.
+             */
+            if (exception instanceof UnknownHostException) {
```
torbsto, line 65: Did you test this? Is this actually the exception that is thrown in such cases?

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +35,61 @@
...
+                final String keyInfo = Arrays.stream(url.split("/")).skip(3).collect(Collectors.joining("/"));
```
torbsto, line 66: You can just use OkHttp's `HttpUrl` class to retrive this. You can get the current one by calling `request.httpUrl()`

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +35,61 @@
...
+                log.info("Host at {} is unavailable. Forwarding the request to {}", url,
+                    fallbackServiceHost);
+                final String newUrl = this.fallbackServiceHost.concat("/".concat(keyInfo));
+                final Request fallbackRequest = new Request.Builder().url(newUrl).get().build();
+                try {
+                    final Response fallbackResponse = this.client.newCall(fallbackRequest).execute();
+                    fallbackResponse.header(HeaderConstants.getCacheMissHeaderName(),
```
torbsto, line 73: It feels wrong to me to mutate the original response to propagate this information. I'd create an additional method in `ResponseWrapper` to cover this case.

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+import okhttp3.ResponseBody;
+
+/**
+ * A wrapper for a response from a call to be made with the HttpClient.
+ *
+ * <p>
+ * It consists of a response body extracted from the response and a boolean that indicates
+ * whether the X-Cache-Update header has been set. This header signals the need to update
+ * the mapping between partitions and mirror hosts.
+ * </p>
```
torbsto, line 34:
```suggestion
 *
```

JavaDoc is weird, there is actually no need to close the tags

torbsto (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
DawidNiezgodka: If you run the following test: `PartitionedMirrorClientTest.shouldReadHeaderAndUpdateRouterWithFewerReplicas`, you will eventually hit this `if` statement. With the host:` http://localhost:1234/mirror/dummy`, which points nowhere, you receive: `java.net.UnknownHostException: local: nodename nor servname provided, or not known`.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
DawidNiezgodka: Wait, but you're not changing the value of any variable here. There is a new request and a new response. Where can you see a mutation of the original response here?

DawidNiezgodka: The method itself can be put into ResponseWrapper.

torbsto: Sorry, not the original but fallback response

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
torbsto: I'm not sure if 1) this is also the case in a cluster and 2) we want to rely on it. I think it's fine to use it for every IOException. Eventually, we can look into more nuanced recovery strategies.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -16,14 +16,15 @@
...
+import com.bakdata.quick.common.api.model.mirror.MirrorHost;
 import com.bakdata.quick.common.exception.MirrorException;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import io.micronaut.http.HttpStatus;
 import java.io.IOException;
 import lombok.extern.slf4j.Slf4j;
 import okhttp3.Request;
 import okhttp3.Response;
-import okhttp3.ResponseBody;
+import org.jetbrains.annotations.NotNull;
```
torbsto, line 27: No need for `@NotNull`

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +33,56 @@
...
-    public ResponseBody makeRequest(final String url) {
-        final Request request = new Request.Builder().url(url).get().build();
-        /*
-        The reason why the classic try-catch statement and not the try-with-resources is used here,
-        is the fact that the try-with-resources implicitly closes the processed resource in an automatically
-        added (and not visible) finally block. This means that if we used it here, we wouldn't be able
-        to read from the InputStream in the caller as it would have been already closed by the try-with-resources
-        in the callee.
-         */
+    // This code covers a situation where a replica is removed, and we can no longer
```
torbsto, line 58: JavaDoc for methods

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -32,54 +33,56 @@
...
+    // reach the host for a given partition. The k8s-service (Load Balancer) is used
+    // as a fallback in such a case. It might also happen that the Load Balancer itself is not reachable.
+    // If this occurs, we throw an exception. We also set the X-Cache-Update header to a response
+    // to immediately update the mapping.
+    @NotNull
+    private ResponseWrapper getResponseFromFallbackService(final String url, final Request request) {
```
torbsto, line 64: Why is this part of `DefaultMirrorRequestManager`? Isn't that only relevant for the `PartitionedMirrorClient`?

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,117 @@
...
+        final ResponseBody body = getAndCheckResponseBody(fallbackResponse);
+        return new ResponseWrapper(body, true);
+    }
+
+    /**
+     * Extract the ResponseBody from the Response and checks its validity.
+     * @param response the response from a particular service (for example, Mirror or Fallback Service)
+     * @return an instance of ResponseBody or an exception if it could not be retrieved
+     */
+    @NotNull
```
torbsto, line 92: Not needed

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
DawidNiezgodka: We can do sth like this in PartitionedMirrorClient:
    ```
private ResponseWrapper getResponseFromFallbackService(final String initialUrl, final String keyInfo) {
        final String newUrl = this.fallbackService.concat(keyInfo);
        log.info("Host at {} is unavailable. Forwarding the request to {}", initialUrl, newUrl);
        try {
            return this.requestManager.makeRequest(newUrl, true);
        } catch (final IOException fallbackException) {
            throw new MirrorException("Fallback service error", HttpStatus.INTERNAL_SERVER_ERROR,
                fallbackException);
        }
    }
```

File: common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorClient.java [hunk 45]
```diff
@@ -69,21 +69,24 @@ public DefaultMirrorClient(final MirrorHost mirrorHost, final HttpClient client,
...
             Collections.emptyList());
     }
 
     @Override
     @Nullable
     public List<V> fetchValues(final List<K> keys) {
         final List<String> collect = keys.stream().map(Object::toString).collect(Collectors.toList());
-        return this.mirrorRequestManager.sendRequest(this.host.forKeys(collect), this.parser::deserializeList);
+        final ResponseWrapper response = this.mirrorRequestManager.makeRequest(this.host.forKeys(collect));
+        return this.mirrorRequestManager.processResponse(response, this.parser::deserializeList);
```
raminqaf, line 89: In `fetchAll()` we are returning an empty list. Let's return an empty list here as well:
```Java
return Objects.requireNonNullElse(
            this.mirrorRequestManager.processResponse(response, this.parser::deserializeList),
            Collections.emptyList());
```

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.Objects;
+import java.util.Queue;
+import okhttp3.OkHttpClient;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import org.apache.kafka.common.serialization.Serdes;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@MicronautTest
```
raminqaf, line 44: You don't need the `@MicronoutTest`

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,170 @@
...
+class PartitionedMirrorClientTest {
```
raminqaf, line 45: There are a lot of properties here that you can add the `this` to them

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -38,48 +35,28 @@ public DefaultMirrorRequestManager(final HttpClient client) {
...
+        // as it would have been already closed by the try-with-resources in the callee.
         try {
-            final ResponseBody body = this.makeRequest(url);
-            if (body != null) {
-                return parser.parse(body.byteStream()).getValue();
-            }
-            return null;
+            final Response response = this.client.newCall(request).execute();
+            return ResponseWrapper.fromResponse(response);
         } catch (final IOException exception) {
```
raminqaf, line 47: Please consider catching the `IllegalStateException` here

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -65,16 +65,23 @@ public MirrorHost findHost(final K key) {
...
+            ));
         }
         return this.partitionToMirrorHost.get(partition);
     }
 
     @Override
     public List<MirrorHost> getAllHosts() {
         if (this.partitionToMirrorHost.isEmpty()) {
-            throw new IllegalStateException("Router has not been initialized properly.");
+            throw new IllegalStateException("Partition to MirrorHost mapping is empty.");
```
raminqaf, line 78: Use `MirrorException` instead

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -65,16 +65,23 @@ public MirrorHost findHost(final K key) {
         final int partition =
             this.partitionFinder.getForSerializedKey(serializedKey, this.partitionToMirrorHost.size());
         if (!this.partitionToMirrorHost.containsKey(partition)) {
-            throw new IllegalStateException("Router has not been initialized properly.");
+            throw new IllegalStateException(String.format(
```
raminqaf, line 68: Use `MirrorException` instead

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
```diff
@@ -0,0 +1,96 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An implementation of MirrorRequestManager that handles the case in which a replica is removed
+ * and the corresponding host is no longer available. If this occurs, the fallback service is called.
+ */
+@Slf4j
+public class MirrorRequestManagerWithFallback implements MirrorRequestManager {
```
raminqaf, line 33: Can you test this class?

raminqaf (changes requested):

common/src/main/java/com/bakdata/quick/common/api/client/HeaderConstants.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package com.bakdata.quick.common.api.client;
+
+/**
+ * A collection of constants for managing headers.
+ */
+public class HeaderConstants {
+
+    private static final String UPDATE_PARTITION_HOST_MAPPING_HEADER = "X-Cache-Update";
```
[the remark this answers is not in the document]
DawidNiezgodka: According to [this](https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions) discussion, there is no consensus regarding the naming schema for custom headers. The RFC spec used to recommend the X- naming convention, only to abandon it later. Some developers stick to X-, some use custom names, some say there is no convention, some advise not to use custom headers, etc.
If you don't like it, we can rename it.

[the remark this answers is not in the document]
raminqaf: Okay, nice! Could you please provide a short description of why the X-Cache-Update is set or not set?

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
DawidNiezgodka: The functionality is tested in `PartitionedMirrorClientTest.shouldReadHeaderAndUpdateRouterWithFewerReplicas`.

raminqaf:
@DawidNiezgodka, I'll perform an E2E test on this PR and see if the routing works

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRegistryClient.java
```diff
@@ -104,12 +107,12 @@ public Single<List<TopicData>> getAllTopics() {
...
-                                                                      final HttpClient client) {
-        final KnownTypeResolver<TopicData> typeResolver =
-            new KnownTypeResolver<>(TopicData.class, client.objectMapper());
-        final String serviceName = topicRegistryConfig.getServiceName();
-        return new DefaultMirrorClient<>(serviceName, client, MirrorConfig.directAccess(), typeResolver,
-            new DefaultMirrorRequestManager(client));
+                                                                      final HttpClient client, final String topicName) {
+        final KnownTypeResolver<TopicData> typeResolver = new KnownTypeResolver<>(
+            TopicData.class, client.objectMapper());
+        final MirrorHost mirrorHost = new MirrorHost(topicName, MirrorConfig.directAccess());
```
raminqaf, line 113: You should use the `topicRegistryConfig.getServiceName()` for the `mirrorHost`

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRegistryClient.java
```diff
@@ -104,12 +107,12 @@ public Single<List<TopicData>> getAllTopics() {
     }
 
     private static MirrorClient<String, TopicData> createMirrorClient(final TopicRegistryConfig topicRegistryConfig,
-                                                                      final HttpClient client) {
-        final KnownTypeResolver<TopicData> typeResolver =
-            new KnownTypeResolver<>(TopicData.class, client.objectMapper());
-        final String serviceName = topicRegistryConfig.getServiceName();
-        return new DefaultMirrorClient<>(serviceName, client, MirrorConfig.directAccess(), typeResolver,
-            new DefaultMirrorRequestManager(client));
+                                                                      final HttpClient client, final String topicName) {
```
raminqaf, line 110: You don't need the `topicName` here. Since you have the topic registry name in the `TopicRegsitryConfig` object. Please remove it.

raminqaf (changes requested):
Also the `MirrorDataFetcherClient` in the gateway is still using the `DefaultMirrorClient` and not the `PartitionedMirrorClient`.

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRegistryClient.java
```diff
@@ -105,11 +108,10 @@ public Single<List<TopicData>> getAllTopics() {
...
                                                                       final HttpClient client) {
-        final KnownTypeResolver<TopicData> typeResolver =
-            new KnownTypeResolver<>(TopicData.class, client.objectMapper());
-        final String serviceName = topicRegistryConfig.getServiceName();
-        return new DefaultMirrorClient<>(serviceName, client, MirrorConfig.directAccess(), typeResolver,
-            new DefaultMirrorRequestManager(client));
+        final KnownTypeResolver<TopicData> typeResolver = new KnownTypeResolver<>(
+            TopicData.class, client.objectMapper());
+        final MirrorHost mirrorHost = new MirrorHost(topicRegistryConfig.getServiceName(), MirrorConfig.directAccess());
+        return new PartitionedMirrorClient<>(mirrorHost, client, Serdes.String(), typeResolver, new DefaultPartitionFinder());
```
raminqaf, line 114: This breaks the TopicRegistry Mirror because you send the request to the TopicRegistry Mirror itself. Concretely, after you initialize the `PartitionedMirrorClient` class you call the `makeRequestForPartitionHostMapping` function in the constructor. This method creates an URL (http://internal-topic-registry/stream/partition) and sends a request, which ends in an internal server error, and the TopicRegistry Mirror fails. Here are also the logs:
```log
INFO com.bakdata.quick.common.api.client.PartitionedMirrorClient - Initializing partition router for the mirror at: internal-topic-registry and the topic: internal-topic-registry.

INFO com.bakdata.quick.common.api.client.MirrorRequestManagerWithFallback - Sending request to: http://internal-topic-registry/streams/partitions
...
INFO com.bakdata.quick.common.api.client.PartitionedMirrorClient - Initializing partition router for the mirror at: internal-topic-registry and the topic: internal-topic-registry. Message: Requested resource responded with status code 500
```
I would suggest that we stick to the `DefaultMirrorClient` for the TopicRegistry and continue to concentrate on the Gateway for the partition routing. Please revert all the changes that involve the TopicRegistry Mirror.

raminqaf (changes requested):

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/QueryListArgumentFetcherTest.java
```diff
@@ -83,34 +90,34 @@ void shouldFetchListWhenListArgumentOfTypeString() throws JsonProcessingExceptio
...
+//            new ListArgumentFetcher<>("productId", fetcherClient, isNullable, hasNullableElements);
+//
+//        final Map<String, Object> arguments = Map.of("productId", List.of(1L, 2L));
+//
+//        final DataFetchingEnvironment env = DataFetchingEnvironmentImpl.newDataFetchingEnvironment()
+//            .localContext(arguments).build();
+//
+//        final List<?> actual = listArgumentFetcher.get(env);
+//        assertThat(actual).isEqualTo(List.of(product1, product2));
+//    }
```
raminqaf, line 120: Please either fix this test or `@Disabled` on this test

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,117 @@
...
+        return body;
+    }
+
+    /**
+     * Checks if the X-Cache-Update from {@link HeaderConstants} header has been set.
+     *
+     * @param response a response from the http call
+     * @return a boolean that indicates whether the X-Cache-Update header has been set
+     */
+    private static boolean checkIfCacheMissHeaderSet(final Response response) {
```
torbsto, line 114: Nit:
```suggestion
    private static boolean isCacheMissHeaderSet(final Response response) {
```

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,117 @@
...
+    ResponseBody responseBody;
+    boolean updateCacheHeaderSet;
+
+    private ResponseWrapper(@Nullable final ResponseBody responseBody, final boolean headerSet) {
+        this.responseBody = responseBody;
+        this.updateCacheHeaderSet = headerSet;
+    }
+
+    private ResponseWrapper(@Nullable final ResponseBody responseBody) {
+        this.responseBody = responseBody;
```
torbsto, line 49: change to `this(responseBody, false)`

common/src/main/java/com/bakdata/quick/common/api/client/HeaderConstants.java
```diff
@@ -0,0 +1,43 @@
...
+ *    limitations under the License.
+ */
+
+package com.bakdata.quick.common.api.client;
+
+/**
+ * A collection of constants for managing headers.
+ * The X-Cache-Update header is set when there is a need to update the
+ * partition-host mapping in the {@link com.bakdata.quick.common.api.client.routing.PartitionRouter}.
+ * The need arises when a replica of the mirror is added or removed.
```
torbsto, line 23: This is not really related to the class itself but to its field `UPDATE_PARTITION_HOST_MAPPING_HEADER`

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManager.java
```diff
@@ -17,30 +17,20 @@
...
-import okhttp3.ResponseBody;
 
 /**
  * Responsible for managing requests send to the Mirror.
  */
 public interface MirrorRequestManager {
 
     /**
-     * Responsible for making a request to a specific url and processing the result.
+     * Submits a request and processes the response. It throws an exception in case of various errors.
```
torbsto, line 27: > It throws an exception in case of various errors.

Either try to be more specific or remove it, as it currently doesn't really add any insight

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public <T> T processResponse(final ResponseWrapper responseWrapper, final ParserFunction<T> parser) {
+        return this.delegate.processResponse(responseWrapper, parser);
+    }
+
+    /**
+     * This code covers a situation where a replica is removed, and we can no longer
+     * reach the host for a given partition. The k8s-service (Load Balancer) is used
+     * as a fallback in such a case. It might also happen that the Load Balancer itself is not reachable.
+     * If this occurs, we throw an exception. We also set the X-Cache-Update header to a response
```
torbsto, line 76:
```suggestion
     * If this occurs, we throw an exception. We also set the {@link HeaderConstants#X-Cache-Update} header to a response
```
Or something like this

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -122,29 +125,57 @@ public Single<MirrorValue<V>> get(final String rawKey) {
...
+     * @param listOfResponses a list of HttpResponses obtained from multiple calls to get(key),
+     *                        see getValues for the details
+     * @return MutableHttpResponse, possibly with a Cache-Miss Header set
+     */
+    private MutableHttpResponse<MirrorValue<List<V>>> transformValuesAndCreateHttpResponse(
+        final List<HttpResponse<MirrorValue<V>>> listOfResponses) {
+        final boolean headerSet = listOfResponses.stream()
+            .anyMatch(response -> response.header(HeaderConstants.getCacheMissHeaderName()) != null);
+        final List<V> values = listOfResponses.stream()
+            .map(response -> Objects.requireNonNull(response.body()).getValue())
```
torbsto, line 155: Can we really require non null here? Why not filter out empty bodies?

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
torbsto: There is no need to expose the constants through getters. You can make them public and remove the `get...`.

common/src/main/java/com/bakdata/quick/common/api/client/DefaultMirrorRequestManager.java
```diff
@@ -38,48 +35,28 @@ public DefaultMirrorRequestManager(final HttpClient client) {
...
 
     @Override
-    @Nullable
-    public <T> T sendRequest(final String url, final ParserFunction<T> parser) {
+    public ResponseWrapper makeRequest(final String url) {
+        final Request request = new Request.Builder().url(url).get().build();
+        // The classic try-catch statement and not the try-with-resources is used here because the try-with-resources
+        // implicitly closes the processed resource in an automatically added (and not visible) block.
+        // If we used it here, we wouldn't be able to read from the InputStream in the caller
+        // as it would have been already closed by the try-with-resources in the callee.
```
torbsto, line 43: Simplify to something like:
> Don't close response here because its content is read later

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
```diff
@@ -0,0 +1,96 @@
...
+    public MirrorRequestManagerWithFallback(final HttpClient client, final MirrorHost fallbackServiceHost) {
+        this.client = client;
+        this.fallbackServiceHost = fallbackServiceHost.plainUrl();
+        this.delegate = new DefaultMirrorRequestManager(client);
+    }
+
+    @Override
+    public ResponseWrapper makeRequest(final String url) {
+        final Request request = new Request.Builder().url(url).get().build();
+        // The classic try-catch statement and not the try-with-resources is used here because the try-with-resources
```
torbsto, line 54: see above

common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java
```diff
@@ -0,0 +1,96 @@
...
+    private ResponseWrapper getResponseFromFallbackService(final String initialUrl, final Request initialRequest) {
+        final String keyInfo = String.join("/", initialRequest.url().pathSegments());
+        final String newUrl = this.fallbackServiceHost + keyInfo;
+        log.info("Host at {} is unavailable. Forwarding the request to {}", initialUrl, newUrl);
+        final Request fallbackRequest = new Request.Builder().url(newUrl).get().build();
+        try {
+            final Response fallbackResponse = this.client.newCall(fallbackRequest).execute();
+            return ResponseWrapper.fromFallbackResponse(fallbackResponse);
+        } catch (final IOException fallbackException) {
+            throw new MirrorException("Fallback service error", HttpStatus.INTERNAL_SERVER_ERROR,
```
torbsto, line 92: This is a user-facing error, therefore it should be more meaningful and descriptive. Try to imagine what you would want to know in such a situation and answer those questions. Is there something the user can do? Is it their fault?

common/src/main/java/com/bakdata/quick/common/api/client/MirrorValueParser.java
```diff
@@ -55,6 +62,13 @@ public MirrorValue<V> deserialize(final InputStream inputStream) throws IOExcept
         return new MirrorValue<>(this.parseValue(value));
     }
 
+    /**
+     * A similar functionality as to the deserialize function but adjusted to work with lists.
```
torbsto, line 66: If you reference the other function, please link it with `{@link ...}`

common/src/main/java/com/bakdata/quick/common/api/client/ResponseWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        final ResponseBody body = getAndCheckResponseBody(response);
+        if (response.header(HeaderConstants.getCacheMissHeaderName()) != null) {
+            return new ResponseWrapper(body, isCacheMissHeaderSet(response));
+        }
+        return new ResponseWrapper(body);
+    }
+
+    /**
+     * Similar to fromResponse, it creates instances of ResponseWrapper. The difference is
```
torbsto, line 70: see above; link to method

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private static final String DEFAULT_TOPIC = "dummy";
+
+    private final MockWebServer server = new MockWebServer();
+    private final ObjectMapper mapper = new ObjectMapper();
+    private final HttpClient client = new HttpClient(this.mapper, new OkHttpClient());
+    private final String host = String.format("%s:%d", this.server.getHostName(), this.server.getPort());
+    private final MirrorHost mirrorHost = new MirrorHost(this.host, MirrorConfig.directAccess());
+    private MirrorClient<String, TopicData> topicDataClient;
```
torbsto, line 53: Since you've switched back to the DefaultMirrorClient for the registry, maybe it would be better to use a different value type here to avoid confusion. What do you think?

common/src/test/java/com/bakdata/quick/common/api/client/PartitionedMirrorClientTest.java
```diff
@@ -0,0 +1,167 @@
...
+        this.partitionFinder.enqueue(1);
+        final TopicData topic2 = this.topicDataClient.fetchValue(DEFAULT_TOPIC);
+        // If the info had not been updated, we would have received IllegalStateException because
+        // the partitionToMirrorHost would not have access to the partition=3.
+        assertThat(Objects.requireNonNull(topic).getName()).isEqualTo(DEFAULT_TOPIC);
+        assertThat(Objects.requireNonNull(topic2).getName()).isEqualTo(DEFAULT_TOPIC);
+
+        this.partitionFinder.enqueue(2);
+        // Since there was no mapping update, an exception will be thrown.
+        // There is no host for partition=3
```
torbsto, line 160: Why partition 3? Isn't this about partition 2?

```suggestion
        // There is no host for partition=2
```

No?

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -36,7 +39,9 @@
     private final String topic;
     private final Serde<K> keySerde;
     private final PartitionFinder partitionFinder;
-    private final Map<Integer, MirrorHost> partitionToMirrorHost;
+    private Map<Integer, MirrorHost> partitionToMirrorHost;
+    private final Set<String> distinctHosts;
```
torbsto, line 43: I think this can be converted to a local variable

torbsto (approved):

raminqaf (approved):

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
DawidNiezgodka: `distinctHosts` or `partitionToMirrorHost`? There are both used in more than one place. Generally, IntelliJ is sensitive to variables that could be local and notices you about them. In this case, there is no info about a possible transformation.

torbsto: `distinctHosts` is only used to make the hosts distinct and could be moved to `findDistinctHosts`

Issue #49, opened 2022-07-20
Gateway should not work on individual data types, just on JSON.


DawidNiezgodka:
Consider the scenario in which a user wants to make a query according to the following:
```
type Query {
getPurchase(id: String): Purchase @topic(name: "purchase", keyArgument: "purchaseId")
}
```
with a concrete query being:
```
{
getPurchase(id: "abc")  {
    productId
    }
}
```

What happens behind the scenes, for example in `QueryKeyArgumentFetcher.get()`:
1. We retrieve the argument ("abc") from the DataFetchingEnvironment, which is JSON representation of the graphql schema,
2. We fetch a value of a specific type (Double, Avro, Protobuf) from the `DataFetcherClient`, 
3. This value is then converted by the framework (which one?) to JSON and delivered to the client. 

If we start and end up with JSON, we make unnecessary conversions. Thus, we might refrain from working with different data types (f.e., Double, Protobuf, Avro) in Gateway and work directly with JSON.

For this:
a) `MirrorDataFetcherClient` has to be rebuilt so that it does not work with `TypeResolver<V>` but with JSON,
b) `MirrorClient` should not receive a resolver that works on a given data type but simply on JSON.

Additionally, it might be considered to complete remove the generic V parameter from Gateway.

DawidNiezgodka:
Changes made in this issue do not involve `SubsciptionFetcher` and `MultiSubscriptionFetcher`.

PR #52, opened 2022-08-03
Leave generic types in Gateway (where possible) and use JSON instead

DawidNiezgodka:
A PR for #49.

Current problems are:

- [x] KeyFieldFetcher returns an Object that can be both a list of JsonNodes or a JsonNode - maybe we can use an ArrayNode there, but I don't know if we can later access its elements if we cast back to JsonNode,
- [x] There is a problem with marshaling to the JsonNode when the target is a String,
- [x] Fix the problem with The `QueryKeyArgumentFetcherTest.shouldFetchLongValue` test. It fails because the original `LongNode` is later cast to the `IntNode`. Strangely, the problem does not arise with `double` and `DoubleNode`,
EDIT: The reason is the following: A long is converted from Java's `long`, i.e., `long a = 5L`;, into JSON, and then again into Java. However, for JSON it is the same, i.e. there is no differentiation between an int and a long.
**Additional question**: Is it problematic for GraphQL if it expects a long but receives an int? Probably not as it doesn't have any concept of long.
- [x] Check if MirrorValueParser can work with `this.resolver.fromString(element.toString())` only,
Preliminary yes.
- [x] Fix the serialization error in `QuickPropertyDataFetcher.get()`,
- [x] Check if `QuickPropertyDataFetcher` can be dropped (as we no longer work with Avro or Proto there),
- [x] Immediately convert to JSON in SubscriptionFetcher,
- [x] Solve the problem with double quotes - _a custom serializer is provided, but it has not been tested yet_,
- [x] Check if a solution with the generic DataFetcher that works as it did for the primitive types and only uses JsonNode for Avro and Proto makes sense, i.e., if it is maintainable and extensible.


gateway/src/main/java/com/bakdata/quick/gateway/custom/QuickPropertyDataFetcher.java
```diff
@@ -34,43 +33,37 @@
...
-        }
-
-        if (environment.getSource() instanceof Message) {
-            return this.extractFromMessage(environment);
-        }
-        return this.propertyDataFetcher.get(environment);
+    public Object get(final DataFetchingEnvironment environment) {
+        final JsonNode source = environment.getSource();
+        final JsonNode value = source.get(fieldName);
+        return extractValueFromJsonNode(value);
```
torbsto, line 51: Either make `extractValueFromJsonNode` static or prefix with `this.`

gateway/src/main/java/com/bakdata/quick/gateway/custom/QuickPropertyDataFetcher.java
```diff
@@ -34,43 +33,37 @@
  * It first checks whether the field returned by the parent is of type {@link GenericRecord}.
```
torbsto, line 33: Docs need to be updated

gateway/src/main/java/com/bakdata/quick/gateway/custom/QuickPropertyDataFetcher.java
```diff
@@ -34,43 +33,37 @@
...
-    }
-
-    // save casting because type safety should be enforced by the GraphQL schema
-    @SuppressWarnings("unchecked")
-    private T extractFieldFromRecord(final GenericRecord genericRecord) {
-        return (T) genericRecord.get(this.fieldName);
+    private Object extractValueFromJsonNode(final JsonNode value) {
+        if (value.isInt()) {
+            return value.asInt();
+        } else if (value.isObject()) {
```
torbsto, line 57: You could merge isObject and isArray

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/DataFetcherClient.java
```diff
@@ -30,6 +30,7 @@
     TypeReference<List<Map<String, Object>>> LIST_TYPE_REFERENCE = new TypeReference<>() {
     };
 
+
```
torbsto, line 33: remove

gateway/src/test/java/com/bakdata/quick/gateway/GraphQLQueryExecutionTest.java
```diff
@@ -87,7 +88,7 @@ void shouldExecuteDefinitionWithSingleFieldAndObjectName(final TestInfo testInfo
         final Map<String, Map<String, Object>> data = executionResult.getData();
         assertThat(data.get("getURL"))
             .isNotNull()
-            .containsEntry("url", "test-url");
+            .containsEntry("url", "\"test-url\"");
```
torbsto, line 91: Why is this still in quotes?

DawidNiezgodka: If I understand correctly, you can't get rid of double quotes at this point. The execution logic doesn't use the `GraphQLJsonSerializer`, so there is no way of removing the double quotes. If you have `TextNode.valueOf("url")`, you will get double quotes. What you can do about it is to remove the quotes when you want to pass the value further. In the `CustomGraphQLJsonSerializer`, we have a piece of code that handles this: 
```
            if (object instanceof TextNode) {
                return ((TextNode) object).textValue();
            } else {
                return objectMapper.writeValueAsString(object);
            }
```
In case of string (TextNode), you will receive a value without double quotes.

torbsto: But where does the conversion from JsonNode to Map<String, Object>> happen then?

torbsto:
>  Immediately convert to JSON in SubscriptionFetcher,

This isn't covered yet, is it?

gateway/src/test/java/com/bakdata/quick/gateway/GraphQLQueryExecutionTest.java
DawidNiezgodka: Unfortunately, I can't set a breakpoint there but in the ExecutionStrategy there is: `serialized = scalarType.getCoercing().serialize(result);`, which is (probably) responsible for fetching the value. I don't know if we have a possibility to interfere with it.

gateway/src/main/java/com/bakdata/quick/gateway/subscriptions/CustomGraphQLJsonSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.IOException;
+
+/**
+ * CustomGraphQLJsonSerializer mitigates the problem of returning double-quoted strings from the gateway.
+ * This problem arises when a fetcher retrieves a string from a mirror, which is later
+ * converted to TextNode. With the default serializer, we obtain a double-quoted string due to serialisation.
+ * However, we want to return a simple string to the user, so for strings (TextNodes), we must extract a text value
+ * instead of rewriting it as a string, which results in a double-quoted string.
+ */
+public class CustomGraphQLJsonSerializer implements GraphQLJsonSerializer {
```
raminqaf, line 31: Do we still need this?

gateway/src/main/java/com/bakdata/quick/gateway/JsonValue.java
```diff
@@ -0,0 +1,85 @@
...
+                return this.jsonNode.asInt();
+            } else if (this.jsonNode.isDouble()) {
+                return this.jsonNode.asDouble();
+            } else if (this.jsonNode.isLong()) {
+                return this.jsonNode.asLong();
+            } else if (this.jsonNode.isBoolean()) {
+                return this.jsonNode.asBoolean();
+            } else if (this.jsonNode.isArray() || this.jsonNode.isObject()) {
+                return this.jsonNode;
+            } else {
```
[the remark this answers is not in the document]
DawidNiezgodka: Here, there is a question if we want to return null when there are null elements. If we throw the BadArgumentException instead of returning null, we will never get the null from this function, which makes the hasNullableElements flag unnecessary.

DawidNiezgodka:
I am closing, for now, not to thwart the implementation of range queries. The issue will be revisited later to solve the problems with `SubscritpionFetcher` and `MultiSubscriptionFetcher`.

PR #70, opened 2022-09-07
Fail a build if it does not meet the checkstyle requirements

### LINKED ISSUE #64: Enforce checkstyle in the CI
We use [checkstyle](https://checkstyle.sourceforge.io/) for our coding style. But currently, we are not enforcing the check in our CI runs. We should add this check a let the CI fail.

DawidNiezgodka:
This resolves #64, however it proposes a different approach. Instead of creating a new GitHub Action, it changes the settings of the checkstyle plugin so that it does not permit any error and warnings.


torbsto:
Didn't you show me that there were a couple of warnings? Why does the CI pass then?

.github/actions/test-source/action.yml
```diff
@@ -23,10 +23,14 @@ runs:
         restore-keys: |
           ${{ runner.os }}-gradle-
 
+    - name: Run build to check if there are any checkstyle issues
```
torbsto, line 26:
```suggestion
    - name: Build project
```

This will also result in a fail if there are any compilation errors

.github/actions/test-source/action.yml
```diff
@@ -23,10 +23,14 @@ runs:
...
+      shell: bash
+      run: ./gradlew build
+    
     - name: Run unit tests
       shell: bash
       run: ./gradlew unitTest
-
+
```
torbsto, line 33: Revert

torbsto (approved):
Thanks :+1:

PR #77, opened 2022-09-16
Fix and refactor stream state host and partition routing

### LINKED ISSUE #80: Partition routing not working properly
### Issue description

The partition routing in the gateway is not working properly for mirrors with range index. There are some bugs and problems I noticed while working with the partition routing more in-depth:

1. The [StreamsStateHost](https://github.com/bakdata/quick/blob/d0c1cb5d49f1755f93ec68110d80700a40b5b171/common/src/main/java/com/bakdata/quick/common/api/client/StreamsStateHost.java#L58) creates a wrong URL. So the request never arrives at the Mirror. 

2. The [getResponseFromFallbackService](https://github.com/bakdata/quick/blob/c8778ce527575c545a864ccbc3d98e3502fbb2a2/common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java#L81) method in [MirrorRequestManagerWithFallback](https://github.com/bakdata/quick/blob/c8778ce527575c545a864ccbc3d98e3502fbb2a2/common/src/main/java/com/bakdata/quick/common/api/client/MirrorRequestManagerWithFallback.java) creates the wrong URL because it is ignoring the query parameters of the initial URL.

3. The [MirrorDataFetcherClient creates a PartitionedMirrorClient using the String SerDe](https://github.com/bakdata/quick/blob/0e09aaba24e7f9b1d976a16968930dfc98069d21/gateway/src/main/java/com/bakdata/quick/gateway/fetcher/MirrorDataFetcherClient.java#L91). This causes that the partition calculation gets wrong. We need to know the type of the key that user is sending with its query request.

### Expected Behavior

That the partition router updates the partition to host map correctly.

### Current Behavior

The partition router in the gateway fails to fetch the information from a mirror and cannot access the correct mirror.

### Steps to Reproduce

1. Setup a gateway, mirror with range index
2. send the first request to register the partition to host id
3. kill the mirror to get a new ip
4. send the request again

raminqaf:
Fixes #80


common/src/main/java/com/bakdata/quick/common/api/client/ClientUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Creates a new URL from a {@link Request} object.
+     *
+     * @param request The request object containing the URL
+     * @param mirrorHost The Mirror host which contains the new URL
+     */
+    public static String createMirrorUrlFromRequest(final Request request, final MirrorHost mirrorHost) {
+        String keyInfo = String.join("/", request.url().pathSegments());
```
torbsto, line 36: Couldn't this be written as something like: `request.newBuilder().setHost(mirrorHost.host()).build().toString()`

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -116,21 +121,50 @@ public List<V> fetchAll() {
...
     public List<V> fetchValues(final List<K> keys) {
-        log.debug("Fetching values for {} keys.", keys.size());
-        return keys.stream().map(this::fetchValue).collect(Collectors.toList());
+        log.debug("Fetching values for keys {}.", keys.size());
+        final List<V> valuesFromAllHosts = new ArrayList<>();
+
+        final Map<MirrorHost, List<K>> mirrorHostKeyMap = new HashMap<>();
+        for (final K key : keys) {
+            final MirrorHost mirrorHost = this.router.findHost(key);
+            if (mirrorHostKeyMap.containsKey(mirrorHost)) {
```
torbsto, line 130: Either use `computeIfAbsent(key, key -> new ArrayList()).add(key)` or extract common `mirrorHostKeyMap.get(mirrorHost).add(key)`

common/src/main/java/com/bakdata/quick/common/api/client/PartitionedMirrorClient.java
```diff
@@ -116,21 +121,50 @@ public List<V> fetchAll() {
...
+            } else {
+                mirrorHostKeyMap.put(mirrorHost, new ArrayList<>());
+                mirrorHostKeyMap.get(mirrorHost).add(key);
+            }
+        }
+        log.debug("Created a map of host and list of keys: {}", mirrorHostKeyMap);
+
+        for (final Entry<MirrorHost, List<K>> mirrorHostWitKeys : mirrorHostKeyMap.entrySet()) {
+            final List<String> stringKeys = mirrorHostWitKeys.getValue().stream().map(Objects::toString)
+                .collect(Collectors.toList());
```
torbsto, line 141:
```suggestion
            final List<String> stringKeys = mirrorHostWitKeys.getValue().stream()
                .map(Objects::toString)
                .collect(Collectors.toList());
```

common/src/main/java/com/bakdata/quick/common/api/client/routing/PartitionRouter.java
```diff
@@ -95,7 +84,21 @@ public List<MirrorHost> getAllHosts() {
...
+        final Set<String> distinctHosts = new HashSet<>(this.partitionToMirrorHost.size());
+        return this.partitionToMirrorHost.values()
+            .stream()
+            .filter(mirrorHost -> distinctHosts.add(mirrorHost.getHost()))
+            .collect(Collectors.toList());
+    }
+
+    private static Map<Integer, MirrorHost> convertHostStringToMirrorHost(final Map<Integer, String> partitionToHost) {
+        return partitionToHost.entrySet().stream().collect(
+            Collectors.toMap(Map.Entry::getKey, e -> new MirrorHost(e.getValue(), MirrorConfig.directAccess())));
```
torbsto, line 102:
```suggestion
            Collectors.toMap(Map.Entry::getKey, entry -> new MirrorHost(entry.getValue(), MirrorConfig.directAccess())));
```

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/FetcherFactory.java
```diff
@@ -121,41 +122,33 @@ public <K, V> DataFetcher<V> mutationFetcher(final String topic, final String ke
...
-    private <K, V> Lazy<QuickTopicData<K, V>> getTopicData(final String topic) {
-        return new Lazy<>(() -> {
-            final Single<QuickTopicData<K, V>> topicData = this.topicTypeService.getTopicData(topic);
-            return topicData.blockingGet();
-        });
-    }
-
-    private <K, V> Lazy<QuickTopicData<K, V>> getTopicDataWithTopicTypeService(final String topic) {
+    private Lazy<QuickTopicData<K, V>> getTopicData(final String topic) {
+        log.debug("Requesting topic data from topic {}", topic);
```
torbsto, line 148: Why did you move the log statement out? The request happens only when the supplier is called

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/MirrorDataFetcherClient.java
```diff
@@ -79,16 +73,17 @@ public List<V> fetchList() {
...
+    private PartitionedMirrorClient<K, V> createMirrorClient(final String host,
+        final MirrorConfig mirrorConfig,
+        final HttpClient client,
+        final QuickTopicData<K, V> quickTopicData) {
         final MirrorHost mirrorHost = new MirrorHost(host, mirrorConfig);
-        return new PartitionedMirrorClient<>(mirrorHost, client, Serdes.String(),
-            valueResolver, new DefaultPartitionFinder());
+
+        log.info("Creating a partitioned mirror client with with service {}", mirrorHost);
+        return new PartitionedMirrorClient<>(mirrorHost, client, quickTopicData, new DefaultPartitionFinder());
```
torbsto, line 87: Like we've talked about, I think this should be done through dependency injection

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/DeferFetcher.java
```diff
@@ -66,12 +66,11 @@ public class DeferFetcher implements DataFetcher<DataFetcherResult<Object>> {
      * @param environment current data fetching environment
      * @return object if argument was found, empty otherwise
      */
-    public static Optional<Object> getArgument(final String argument, final DataFetchingEnvironment environment) {
+    public Optional<T> getArgument(final String argument, final DataFetchingEnvironment environment) {
```
torbsto, line 69: Why is this no longer static?

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/ListArgumentFetcher.java
```diff
@@ -71,13 +72,13 @@ public ListArgumentFetcher(final String argument,
     @Override
     @Nullable
     public List<V> get(final DataFetchingEnvironment environment) {
-        final Object arguments = DeferFetcher.getArgument(this.argument, environment)
+        final DeferFetcher<K> deferFetcher = new DeferFetcher<>();
```
torbsto, line 75: see above; I don't see the reason why this isn't static anymore. Also, if you really want this to be an object, why not move the creation into the constructor?

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/subscription/MultiSubscriptionFetcher.java
```diff
@@ -46,10 +46,10 @@
  * data for the other selected fields is fetched through the mirror.
  */
 @Slf4j
-public class MultiSubscriptionFetcher implements DataFetcher<Publisher<Map<String, Object>>> {
+public class MultiSubscriptionFetcher<K> implements DataFetcher<Publisher<Map<String, Object>>> {
 
     public static final int CACHE_SIZE = 5_000;
-    private final Map<String, DataFetcherClient<?>> fieldDataFetcherClients;
+    private final Map<String, DataFetcherClient<String, ?>> fieldDataFetcherClients;
```
torbsto, line 52: Why String? We don't know the type here, do we?

gateway/src/main/java/com/bakdata/quick/gateway/fetcher/subscription/MultiSubscriptionFetcher.java
```diff
@@ -166,7 +166,7 @@ private Mono<Map<String, Object>> createComplexType(final NamedRecord<?, ?> name
 
     @Nullable
     private Object loadField(final FieldKey<?> fieldKey) {
-        final DataFetcherClient<?> client = this.fieldDataFetcherClients.get(fieldKey.getFieldName());
+        final DataFetcherClient<String, ?> client = this.fieldDataFetcherClients.get(fieldKey.getFieldName());
         Objects.requireNonNull(client, () -> "No client found for field " + fieldKey.getFieldName());
         return client.fetchResult(fieldKey.getKey().toString());
```
torbsto, line 171: This will fail if the key serde is for an integer, no?

mirror/src/main/java/com/bakdata/quick/mirror/service/KafkaQueryService.java
```diff
@@ -135,6 +135,17 @@ public Single<HttpResponse<MirrorValue<List<V>>>> getValues(final List<String> k
...
 
+    @Override
+    public Single<HttpResponse<MirrorValue<List<V>>>> getAll() {
+        // For now, we only consider the local state!
+        final ReadOnlyKeyValueStore<K, V> store = this.streams.store(this.storeQueryParameters);
+        return Flowable.fromIterable(store::all)
+            .map(keyValue -> keyValue.value)
+            .toList()
+            .map(valuesList -> HttpResponse.created(new MirrorValue<>(valuesList)).status(200));
+
```
torbsto, line 146: Remove line

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/FetcherTest.java
```diff
@@ -0,0 +1,63 @@
...
+ *    limitations under the License.
+ */
+
+package com.bakdata.quick.gateway.fetcher;
+
+import java.util.List;
+import lombok.Builder;
+import lombok.Value;
+
+public abstract class FetcherTest {
```
torbsto, line 23: I don't see the point of having this as a base class?

gateway/src/test/java/com/bakdata/quick/gateway/fetcher/KeyFieldFetcherTest.java
```diff
@@ -64,126 +51,87 @@ void shouldFetchModificationValue() throws Exception {
...
     }
 
-    @Data
-    @Builder
-    private static class Currency {
-        private String currencyId;
-        private String currency;
-        private double rate;
-    }
+//    @Test
```
torbsto, line 68: Still commented

gradle.properties
```diff
@@ -1,4 +1,4 @@
-org.gradle.caching=true
+org.gradle.caching=false
```
torbsto, line 1: Are you sure this is required?

raminqaf: I always have to change this because the generated Avro or Proto code are somehow cached and they keep breaking the build

torbsto (approved):

common/src/main/java/com/bakdata/quick/common/api/client/mirror/StreamsStateHost.java
```diff
@@ -14,29 +14,31 @@
...
+public final class StreamsStateHost {
+    private static final String DEFAULT_MIRROR_SCHEME = "http";
     private final String host;
-    private final MirrorConfig config;
 
     /**
      * Private constructor for creating StreamsStateHost.
      *
-     * @param host   the host of the mirror. This can be a service name or an IP.
+     * @param topic the host of the mirror. This can be a service name or an IP.
```
DawidNiezgodka, line 37: "topic the host of the mirror" does not seem right. Should it be "the topic of the mirror"?

common/src/main/java/com/bakdata/quick/common/api/model/mirror/MirrorHost.java
```diff
@@ -17,71 +17,137 @@
...
+    private static final String DEFAULT_MIRROR_SCHEME = "http";
+    private final String topic;
     private final MirrorConfig config;
+    private final String host;
 
     /**
      * Default constructor.
      *
-     * @param host the host of the mirror. This can be a service name or an IP.
+     * @param topic the host of the mirror. This can be a service name or an IP.
```
DawidNiezgodka, line 41: same as before

DawidNiezgodka (approved):

Issue #73, opened 2022-09-09
Check for `--range-field` and `--retention-time` during deployment


raminqaf:
The `--retention-time` option should only be set with `--point` option and not with the `--range-field` option

DawidNiezgodka:
Resolved by https://github.com/bakdata/quick/pull/84.

raminqaf:
related to #55

PR #84, opened 2022-09-21
Removed point/no-point, fixed retention-time - range-field exclusion

### LINKED ISSUE #83: Adjust manager to the change in the --point / --no-point approach
We decided to drop the possibility of not creating a mirror index. A mirror index is always created by default. Thus, there is no need for `--point` or `--no-point` flags. Since Manager has already been adjusted to these flags, it must re-adjusted again.

DawidNiezgodka:
Resolves #83 and #73.

### DIFF
--- a/manager/src/main/java/com/bakdata/quick/manager/mirror/resources/MirrorResourceLoader.java
+++ b/manager/src/main/java/com/bakdata/quick/manager/mirror/resources/MirrorResourceLoader.java
@@ -167,8 +167,8 @@ public class MirrorResourceLoader implements ResourceLoader<MirrorResources, Mir
             .put("--input-topics", topic);
 
         if (Objects.nonNull(retentionTime) && Objects.nonNull(rangeField)) {
-            throw new BadArgumentException("The --range-field option must not be specified"
-                + " when --retention-time is set");
+            final String errorMessage = "The --range-field option must not be specified when --retention-time is set";
+            throw new BadArgumentException(errorMessage);
         }
 
         if (Objects.nonNull(retentionTime)) {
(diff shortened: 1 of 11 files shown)


raminqaf (changes requested):
Please update the open API specs and generate them for the PR https://github.com/bakdata/quick-cli/pull/10

raminqaf (approved):
Looks good. Here the documentation how you can create the client code for the CLI
https://bakdata.github.io/quick/dev/developer/development/#openapi

manager/src/main/java/com/bakdata/quick/manager/mirror/resources/MirrorResourceLoader.java
```diff
@@ -163,23 +162,21 @@ private Service createMirrorService(final String name) {
...
     private static Map<String, String> createCliArguments(final String topic,
         @Nullable final Duration retentionTime,
-        final boolean point,
         @Nullable final String rangeField) {
         final ImmutableMap.Builder<String, String> builder = ImmutableMap.<String, String>builder()
             .put("--input-topics", topic);
 
-        builder.put("--point", Boolean.toString(point));
+        if (Objects.nonNull(retentionTime) && Objects.nonNull(rangeField)) {
+            throw new BadArgumentException("The --range-field option must not be specified"
```
raminqaf, line 170: Extract the error message in a variable so it does not break in the line.

raminqaf (approved):

manager/src/main/java/com/bakdata/quick/manager/mirror/resources/MirrorResourceLoader.java
DawidNiezgodka: Done. But is it really that relevant to not break it?

PR #92, opened 2022-10-05
Fix mirror host direct access bug

raminqaf:
A small bug was introduced during the refactoring in #77.
Using the builder of HttpUrl the host should not contain any port number. However, in some cases, we use `directaccess()` in the MirrorHost and use an IP along with its port instead of the service name.


common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorHost.java
```diff
@@ -118,10 +119,10 @@ public HttpUrl forRange(final String key, final String from, final String to) {
...
     @Override
     public String toString() {
-        return new Builder()
-            .scheme(DEFAULT_MIRROR_SCHEME)
-            .host(this.host)
-            .toString();
+        final String url = String.format("%s://%s", DEFAULT_MIRROR_SCHEME, this.host);
+        final HttpUrl parse = HttpUrl.parse(url);
+        return Objects.requireNonNull(parse, "The url is not valid")
+            .newBuilder().toString();
```
torbsto, line 125: Wouldn't `getBaseUrlBuilder.toString()` work?

common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorHost.java
```diff
@@ -145,9 +146,9 @@ public int hashCode() {
     }
 
     private Builder getBaseUrlBuilder() {
-        return new Builder()
-            .scheme(DEFAULT_MIRROR_SCHEME)
-            .host(this.host)
+        final String url = String.format("%s://%s", DEFAULT_MIRROR_SCHEME, this.host);
+        final HttpUrl parse = HttpUrl.parse(url);
```
torbsto, line 150: You could make the HttpUrl a member variable. There is then no need to parse this every time.

common/src/main/java/com/bakdata/quick/common/api/client/mirror/ParserFunction.java
```diff
@@ -17,8 +17,14 @@
 package com.bakdata.quick.common.api.client.mirror;
 
 import com.bakdata.quick.common.api.model.mirror.MirrorValue;
+import com.bakdata.quick.common.config.MirrorConfig;
```
torbsto, line 20: The added imports seem wrong?

raminqaf:
@torbsto, we can add two static factory methods to create the MirrorHost object. One for DirectAccess and one for MirrorConfig. WDYT?

common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorHost.java
```diff
@@ -14,37 +14,54 @@
...
+    /**
+     * Creates the mirror host with the mirrorName. It adds the default {@link MirrorConfig#DEFAULT_MIRROR_HOST_PREFIX}
+     * to the mirror name
+     *
+     * @param mirrorName the name of the Mirror.
+     */
+    public static MirrorHost createMirrorHostWithDefaultPrefix(final String mirrorName) {
+        final MirrorConfig mirrorConfig = new MirrorConfig();
+        final String host = mirrorConfig.getPrefix() + mirrorName;
+        final HttpUrl httpUrl = createUrlFromString(host);
```
torbsto, line 53: Do we still need this here? I thought it was only required because it might have a port , is this ever the case here?

common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorHost.java
```diff
@@ -14,37 +14,54 @@
...
+        this.url = url;
+    }
+
+    /**
+     * Creates the mirror host with the mirrorName. It adds the default {@link MirrorConfig#DEFAULT_MIRROR_HOST_PREFIX}
+     * to the mirror name
+     *
+     * @param mirrorName the name of the Mirror.
+     */
+    public static MirrorHost createMirrorHostWithDefaultPrefix(final String mirrorName) {
```
torbsto, line 50:
```suggestion
    public static MirrorHost createMirrorHostWithPrefix(final String mirrorName) {
```

Might not be the default if configured

torbsto: You could also drop the `MirrorHost` both here and `createMirrorHostWithNoPrefix`. 
`MirrorHost.createWith(No)Prefix` sounds good to me.

e2e/functional/schema/tests.bats
```diff
@@ -28,7 +28,7 @@ setup() {
 }
 
 @test "should create product-topic with schema" {
-    run quick topic create ${TOPIC} --key-type long --value-type schema --schema "${GATEWAY}.${TYPE}"
+    run quick topic create ${TOPIC} --key-type int --value-type schema --schema "${GATEWAY}.${TYPE}"
```
torbsto, line 31: Why did you change this?

common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorHost.java
raminqaf: Yeah, good point! I adjusted it

torbsto (approved):

PR #99, opened 2022-10-07
RQ-Docs adjustment

### LINKED ISSUE #97: Update the range queries documentation with a new example
The current example for range queries is misleading because of the possible override of records. Consider the ingestion of the following two records:
```
  {
    "key": 1,
    "value": {
      "userId": 1,
      "purchaseId": "abc",
      "rating": 7
    }
  }
```
and then:
```
  {
    "key": 1,
    "value": {
      "userId": 1,
      "purchaseId": "def",
      "rating": 7
    }
  }
```
The ingestion of the second leads to the range index for 0[...]1_0[...]7 being overwritten.
We will update the docu with an example of timestamp with the assumption of events that happen continuously and it is not possible to insert the same time for the identical key.

DawidNiezgodka:
This PR resolves #97.


docs/docs/developer/range-query-details.md
```diff
@@ -22,69 +22,79 @@ for arriving at a better understanding of the range queries.
...
-To circumvent the limitation of a key-value store and be able to perform range queries,
+To circumvent the limitation of a key-value store
+and be able to perform range queries,
 Quick uses an alternative approach to deal with keys.
-Each key is a flattened string with a combination of the topic key and the value
+Each key is a flattened string with a combination of the topic key
+and the zero-padded value
 for which the range queries are requested.
-The keys are padded (depending on the type `Int` 10 digits or `Long` 19 digits)
+The values are padded (depending on the type `Int` 10 digits or `Long` 19 digits)
```
raminqaf, line 70:
```suggestion
The values of the range field are padded (depending on the type `Int` 10 digits or `Long` 19 digits)
```

raminqaf (approved):

Issue #100, opened 2022-10-07
Update reference section


raminqaf:
- [x] [Dependency versions](https://bakdata.github.io/quick/dev/user/reference/dependency-versions/)
- [x] [Quick CLI](https://bakdata.github.io/quick/dev/user/reference/cli-commands/)
- [x] [GraphQL support](https://bakdata.github.io/quick/dev/user/reference/graphql-support/)

PR #104, opened 2022-10-11
Refs section update

DawidNiezgodka:
Draft for #100


docs/docs/user/reference/dependency-versions.md
```diff
@@ -3,9 +3,9 @@
...
 |--------------------|---------------------|
 | Quick              | {{ quick_version }} |
-| Kubernetes         | 1.20 - 1.22         |
-| Kafka              | 2.7 - 2.8           |
-| Schema Registry    | 5.5 - 6.2           |
-| Traefik            | 2.4 - 2.5           |
+| Kubernetes         | 1.22                |
+| Kafka              | 2.7 - 3.2           |
+| Schema Registry    | 6.1 - 7.2           |
+| Traefik            | 2.8.4               |
```
raminqaf, line 9: We still support Traefik 2.4

docs/docs/user/reference/dependency-versions.md
```diff
@@ -3,9 +3,9 @@
 | Component          | Version             |
 |--------------------|---------------------|
 | Quick              | {{ quick_version }} |
-| Kubernetes         | 1.20 - 1.22         |
-| Kafka              | 2.7 - 2.8           |
-| Schema Registry    | 5.5 - 6.2           |
-| Traefik            | 2.4 - 2.5           |
+| Kubernetes         | 1.22                |
+| Kafka              | 2.7 - 3.2           |
+| Schema Registry    | 6.1 - 7.2           |
```
raminqaf, line 8: I'm thinking of changing this to the confluent platform version.

torbsto: There are enough people who don't use Confluent platform, so I'd like to keep the Kafka version

PR #106, opened 2022-10-11
Upgrade Kafka to 3.2 and confluent platform to 7.2.2

### LINKED ISSUE #105: Support Kafka 3.2
A follow-up of #5. Currently, we are supporting Kafka 3.1. We should update the dependencies to support 3.2.

raminqaf:
Closes #105


torbsto (commented):
What about the docs? Do we want to change the setup there as well?

raminqaf:
@torbsto Good point. I'll change it to Kafka 3.1 in the docs.

torbsto (approved):

PR #115, opened 2022-10-14
DateTime support

### LINKED ISSUE #110: Support DateTime scalar
Currently, we are not supporting the [DateTime](https://www.the-guild.dev/graphql/scalars/docs/scalars/date-time) scalar type of graphql. This can be useful when working with date and time types.

DawidNiezgodka:
Draft for #110.
EDIT: Needs a further development to involve the change from the comment.


raminqaf (changes requested):
This will not work yet! What happens if the user creates a topic with a GraphQL type with a field of type `DateTime`? The manager needs to convert the schema with `DateTime` type in GraphQL to a DateTime type for Avro and Protobuf.

PR #130, opened 2022-10-21
Refactor mirror range indexer

raminqaf:


mirror/src/main/java/com/bakdata/quick/mirror/range/indexer/RangeIndexer.java
```diff
@@ -14,11 +14,17 @@
...
+package com.bakdata.quick.mirror.range.indexer;
+
+import com.bakdata.quick.mirror.range.MirrorRangeProcessor;
 
 /**
  * Creates the range index in the {@link MirrorRangeProcessor}.
  */
 public interface RangeIndexer<K, V> {
-    String createIndex(final K key, final V value);
+    <F> String createIndex(final K key, final V value);
```
torbsto, line 25: Why `<F>`?

mirror/src/main/java/com/bakdata/quick/mirror/range/indexer/StringValueRangeIndexer.java
```diff
@@ -0,0 +1,75 @@
...
+import io.confluent.kafka.schemaregistry.protobuf.ProtobufSchema;
+import io.micronaut.core.util.StringUtils;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * An indexer for string values. This indexer is used to build the default range index pattern when querying for
+ * ranges.
+ */
+@Slf4j
+public final class StringValueRangeIndexer<K> implements RangeIndexer<K, String> {
```
torbsto, line 37: Could you explain the differences between the default range indexer and this class? I don't understand why this is required

mirror/src/main/java/com/bakdata/quick/mirror/range/indexer/StringValueRangeIndexer.java
```diff
@@ -0,0 +1,75 @@
...
+        this.parsedSchema = parsedSchema;
+        this.rangeField = rangeField;
+    }
+
+    /**
+     * Creates the range index for a given key and a value string type.
+     */
+    public static <K> StringValueRangeIndexer<K> create(final ParsedSchema parsedSchema,
+        final String rangeField) {
+        log.debug("Type Avro detected");
```
torbsto, line 54: (Re)move

mirror/src/main/java/com/bakdata/quick/mirror/range/indexer/RangeIndexer.java
raminqaf: Sry, that shouldn't have been there. I fixed that in the new push.

mirror/src/main/java/com/bakdata/quick/mirror/range/indexer/StringValueRangeIndexer.java
raminqaf: The default range indexer builds the range index when the data is being ingested into the mirror. And the `StringValueRangeIndexer` is used to create the exclusive range index when a query arrives.

mirror/src/main/java/com/bakdata/quick/mirror/range/extractor/type/AvroTypeExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+@Slf4j
+public class AvroTypeExtractor implements FieldTypeExtractor {
+    @Override
+    public QuickTopicType extract(final ParsedSchema parsedSchema, final String fieldName) {
+        final Schema avroSchema = (Schema) parsedSchema.rawSchema();
+        final Schema.Type fieldType = getAvroFieldType(avroSchema, fieldName);
+        log.debug("Field Type is {}", fieldType);
+
+        if (fieldType == Schema.Type.INT) {
```
torbsto, line 40: You could also use a map for this mapping

torbsto (approved):

Issue #119, opened 2022-10-17
Create range index on `--range-key` field in mirror


raminqaf:
The mirror should repartition the data based on the defined **range-key** field.

The mirror should use the [selectKey](https://kafka.apache.org/20/javadoc/org/apache/kafka/streams/kstream/KStream.html#selectKey-org.apache.kafka.streams.kstream.KeyValueMapper-) method to repartition the data based on the new key. Kafka Streams will:

- Send the rekeyed data to an internal repartition topic
- Reread the newly rekeyed data back into Kafka Streams

This issue is related to https://github.com/bakdata/quick/issues/114.

raminqaf:
Closed due to #132

PR #132, opened 2022-10-25
Add support for building range and point index on `range-key` field of the value

raminqaf:
Solves #119


mirror/src/main/java/com/bakdata/quick/mirror/context/RecordData.java
```diff
@@ -14,17 +14,19 @@
...
+ * @param <K> Type of the key
+ * @param <V> Type of the value
  */
-public interface StreamConsumer {
-    /**
-     * Contains the logic of how your stream should be consumed from the input source.
-     */
-    <K, V> KStream<K, V> consume(final MirrorContext<K, V> mirrorContext);
+@Value
+public class RecordData<K, V> {
```
torbsto, line 29: Can we find a name a bit more expressive than `RecordData`? This contains the data for the repartition topic, right? I think the name should reflect that

mirror/src/main/java/com/bakdata/quick/mirror/range/KeySelector.java
```diff
@@ -0,0 +1,79 @@
...
+import com.bakdata.quick.common.type.QuickTopicType;
+import com.bakdata.quick.mirror.range.extractor.type.FieldTypeExtractor;
+import com.bakdata.quick.mirror.range.extractor.value.FieldValueExtractor;
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+import org.apache.kafka.common.serialization.Serde;
+
+/**
+ * Extracts the type and the value from a field in a complex type.
+ */
+public final class KeySelector<R, V> {
```
torbsto, line 32: please document `R` and `V`

mirror/src/main/java/com/bakdata/quick/mirror/range/KeySelector.java
```diff
@@ -0,0 +1,79 @@
...
+        final ConversionProvider conversionProvider,
+        final ParsedSchema parsedSchema,
+        final String rangeKey) {
+        final QuickTopicType keyType = fieldTypeExtractor.extract(parsedSchema, rangeKey);
+        final Class<R> classType = conversionProvider.getClassType(keyType);
+        return new KeySelector<>(keyType, classType, fieldValueExtractor, conversionProvider);
+    }
+
+    /**
+     * extracts the value of a given field.
```
torbsto, line 62:
```suggestion
     * Extracts the value of a given field.
```

mirror/src/main/java/com/bakdata/quick/mirror/MirrorApplication.java
```diff
@@ -127,31 +135,35 @@ public void buildTopology(final StreamsBuilder builder) {
...
-            builder.fieldValueExtractor(new MessageValueExtractor<>());
-        } else if (schemaConfig.getFormat() == SchemaFormat.AVRO) {
-            builder.fieldTypeExtractor(new AvroTypeExtractor());
-            builder.fieldValueExtractor(new GenericRecordValueExtractor<>());
-        }
-
-        this.contextProvider.setMirrorContext(builder.build());
-        return builder.build();
+            .fieldValueExtractor(this.schemaExtractor.getFieldValueExtractor())
+            .fieldTypeExtractor(this.schemaExtractor.getFieldTypeExtractor())
```
torbsto, line 165: You could use just the schema extractor in the context to limit the number of fields there

mirror/src/main/java/com/bakdata/quick/mirror/MirrorApplication.java
```diff
@@ -207,15 +219,11 @@ private static int execute(final ApplicationContext context, final String[] args
...
     protected void runStreamsApplication() {
-        final MirrorContext<K, V> mirrorContext = this.contextProvider.get();
-
-        final MirrorContextBuilder<K, V> builder = mirrorContext.toBuilder()
-            .streams(this.getStreams())
-            .hostInfo(this.hostConfig.toInfo());
-
-        this.contextProvider.setMirrorContext(builder.build());
+        final MirrorContext<?, V> mirrorContext = this.contextProvider.get();
+        final MirrorContext<?, V> mirrorContextWithStream =
```
torbsto, line 223: Why not `R` instead of `?`?

mirror/src/main/java/com/bakdata/quick/mirror/StreamConsumer.java
```diff
@@ -0,0 +1,109 @@
...
+    @Inject
+    public StreamConsumer(final SchemaExtractor schemaExtractor, final ConversionProvider conversionProvider) {
+        this.schemaExtractor = schemaExtractor;
+        this.conversionProvider = conversionProvider;
+    }
+
+    /**
+     * Consumes the input topic and does a repartitioning based on the range key.
+     */
+    // The cast is safe. The generic types of R and K are similar when the rangeKey is null
```
torbsto, line 57:
```suggestion
    // The cast is safe. The generic types of R and K are equal when the rangeKey is null
```

mirror/src/main/java/com/bakdata/quick/mirror/StreamConsumer.java
```diff
@@ -0,0 +1,109 @@
...
+        final RecordData<K, V> recordData =
+            new RecordData<>(topologyData.getTopicData().getKeyData(), topologyData.getTopicData()
+                .getValueData());
+
+        return (RepartitionedTopologyData<R, V>) new RepartitionedTopologyData<>(recordData, stream);
+    }
+
+    private static <K, R, V> KStream<R, V> repartitionStream(final KStream<K, V> inputStream,
+        final KeySelector<R, V> keySelector,
+        final Serde<V> valueSerde, final String rangeKey) {
```
torbsto, line 85:
```suggestion
        final KeySelector<R, V> keySelector, final Serde<V> valueSerde, final String rangeKey) {
```

mirror/src/main/java/com/bakdata/quick/mirror/context/RecordData.java
raminqaf: No, it doesn't only contain the data of the repartition topic. If no repartitioning happens, this object will contain the data of the input topic. That is why I chose a generic name for it.

mirror/src/main/java/com/bakdata/quick/mirror/MirrorApplication.java
```diff
@@ -127,31 +135,34 @@ public void buildTopology(final StreamsBuilder builder) {
...
-            builder.fieldValueExtractor(new MessageValueExtractor<>());
-        } else if (schemaConfig.getFormat() == SchemaFormat.AVRO) {
-            builder.fieldTypeExtractor(new AvroTypeExtractor());
-            builder.fieldValueExtractor(new GenericRecordValueExtractor<>());
-        }
-
-        this.contextProvider.setMirrorContext(builder.build());
-        return builder.build();
+            .schemaExtractor(this.schemaExtractor)
+            .isCleanup(this.cleanUp).build();
```
torbsto, line 165: Nit:
```suggestion
            .isCleanup(this.cleanUp).
            build();
```

mirror/src/main/java/com/bakdata/quick/mirror/StreamConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+            new RecordData<>(topologyData.getTopicData().getKeyData(), topologyData.getTopicData()
+                .getValueData());
+
+        return (RepartitionedTopologyData<R, V>) new RepartitionedTopologyData<>(recordData, stream);
+    }
+
+    private static <K, R, V> KStream<R, V> repartitionStream(final KStream<K, V> inputStream,
+        final KeySelector<R, V> keySelector, final Serde<V> valueSerde, final String rangeKey) {
+        final QuickData<R> repartitionedKeyData = keySelector.getRepartitionedKeyData();
+        final Serde<R> serde = repartitionedKeyData.getSerde();
```
torbsto, line 86:
```suggestion
        final Serde<R> keySerde = repartitionedKeyData.getSerde();
```

mirror/src/main/java/com/bakdata/quick/mirror/context/RecordData.java
torbsto: I still feel like `RecordData` is too generic. Maybe we can find a name that represents both branches of streams, like `IndexInputStream`?

mirror/src/main/java/com/bakdata/quick/mirror/point/MirrorProcessor.java
```diff
@@ -55,6 +57,7 @@ public void process(final Record<K, V> record) {
         if (value == null) {
             this.store.delete(key);
         } else {
+            log.trace("Putting key {} and value {}", key, value);
```
torbsto, line 60: Please add to which store you add it

mirror/src/main/java/com/bakdata/quick/mirror/range/KeySelector.java
```diff
@@ -0,0 +1,82 @@
...
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+import org.apache.kafka.common.serialization.Serde;
+
+/**
+ * Extracts the type and the value from a field in a complex type.
+ *
+ * @param <R> The type of the selected key from the schema
+ * @param <V> The type of the value
+ */
+public final class KeySelector<R, V> {
```
torbsto, line 35: I don't really see why this is called `KeySelector`. It has two tasks: Delegating to the `FieldValueExtractor` to extract the value and creating the data for repartitioned stream. Which task selects the key?

mirror/src/main/java/com/bakdata/quick/mirror/StreamConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.kstream.Consumed;
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.Repartitioned;
+
+/**
+ * Contains the logic of consuming the input topic and repartitioning of the topic.
+ */
+@Singleton
+public class StreamConsumer {
```
torbsto, line 44: I think the name is again a bit too generic, this sounds more like an interface. Maybe (with the comment above in mind) like `IndexInputStreamBuilder`?

mirror/src/main/java/com/bakdata/quick/mirror/StreamConsumer.java
```diff
@@ -0,0 +1,108 @@
...
+    public StreamConsumer(final SchemaExtractor schemaExtractor, final ConversionProvider conversionProvider) {
+        this.schemaExtractor = schemaExtractor;
+        this.conversionProvider = conversionProvider;
+    }
+
+    /**
+     * Consumes the input topic and does a repartitioning based on the range key.
+     */
+    // The cast is safe. The generic types of R and K are equal when the rangeKey is null
+    @SuppressWarnings("unchecked")
```
torbsto, line 58: Could we put the unchecked case into its own method? It's kind of hard to see where it happens right  now

torbsto: I also think a comment here would be helpful to explain the general idea. Why do we need this? How is this done?

torbsto (approved):

PR #138, opened 2022-11-03
Fix keyArgument cast bug

### LINKED ISSUE #137: keyField fetching raises error
### Issue description
On migrating the quick-example profile-store from 0.7.0 to 0.8.0 the keyArgument functionality is broken.
The relevant part of the [schema](https://github.com/bakdata/quick-examples/blob/main/profile-store/deployment/schema.gql) is:
```
type Query {
    getUserProfile(userId: Long!): UserProfile
}

type UserProfile {
    artistCharts: NamedArtistCharts! @topic(name: "lastlisten", keyArgument: "userId")
    ...
}

type NamedArtistCharts {
    topK: [NamedArtistCount!]!
}

type NamedArtistCount {
    id: Long!
    count: Long!
    # here
    artist: Item! @topic(name: "artists", keyField: "id")
}

type Item {
    id: Long!
    name: String!
}
```
The following Query
```
# Q1
query {
 getUserProfile(userId: 21072247) {
   artistCharts {
     topK {
       artist {
        name 
}}}}}
```
causes the following exception in the gateway:
```
WARN notprivacysafe.graphql.execution.SimpleDataFetcherExceptionHandler - Exception while fetching data (/getUserProfile/artistCharts/topK[9]/artist) : class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
    at org.apache.kafka.common.serialization.LongSerializer.serialize(LongSerializer.java:19)
    at com.bakdata.quick.common.api.client.routing.PartitionRouter.findHost(PartitionRouter.java:83)
    at com.bakdata.quick.common.api.client.mirror.PartitionedMirrorClient.fetchValue(PartitionedMirrorClient.java:63)
    at com.bakdata.quick.gateway.fetcher.MirrorDataFetcherClient.fetchResult(MirrorDataFetcherClient.java:44)
    at com.bakdata.quick.gateway.fetcher.KeyFieldFetcher.get(KeyFieldFetcher.java:97)
...
```
### Expected Behavior
Q1 should return an array of objects under the topK-field. The project is set up such that each id has a record in the "artists"-topic.

### Current Behavior
The gateway throws exceptions that are propagated to the query result.

### Steps to Reproduce
Deploy the profile-store example using quick-0.8.0. Execute Q1.

### Detailed Description
* To exclude other causes, notice that the following query works
```
# Q2
query {
 getUserProfile(userId: 21072247) {
   artistCharts {
     topK {
       id
}}}}
```
* Also, all ids obtained with Q2 are present in the artists topic, which can be verified on the mirror endpoint.
In an earlier deployment, the error was null, but this could not be reproduced

raminqaf:
Closes #137


gateway/src/main/java/com/bakdata/quick/gateway/fetcher/KeyFieldFetcher.java
```diff
@@ -149,30 +140,21 @@ private static Object extractCorrectType(final JsonNode jsonNode) {
...
         if (environment.getSource() instanceof GenericRecord) {
             final GenericRecord record = environment.getSource();
-            return new String(this.converter.convertToJson(record), StandardCharsets.UTF_8);
+            return this.objectMapper.readTree(this.converter.convertToJson(record));
         }
 
         if (environment.getSource() instanceof DynamicMessage) {
             final DynamicMessage source = environment.getSource();
-            return JsonFormat.printer().includingDefaultValueFields().print(source);
+            return this.objectMapper.readTree(source.toByteArray());
```
torbsto, line 154: I would be very surprised if this actually works. I guess there is no test case covering protobuf. You need to pass the JSON here.

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -59,4 +76,52 @@ public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveCo
     public boolean isValid(final TopicDirectiveContext context) {
         return context.getTopicDirective().hasKeyField() && !context.isListType();
     }
+
+    /**
+     * Extracts the {@link TypeName} of a  given {@link Type}. The class does not use the default implementation of the
```
torbsto, line 81:
```suggestion
     * Extracts the {@link TypeName} of a given {@link Type}. The class does not use the default implementation of the
```

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -59,4 +76,52 @@ public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveCo
...
+            return (TypeName) type;
+        }
+
+        if (type instanceof NonNullType) {
+            return this.extractName(((NonNullType) type).getType());
+        }
+
+        final String errorMessage = "Found unknown type: " + type.getClass().getSimpleName();
+        log.error(errorMessage);
+        throw new QuickDirectiveException(errorMessage);
```
torbsto, line 96: Please don't log and throw

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -59,4 +76,52 @@ public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveCo
...
+    public TypeName extractName(final Type<?> type) {
+        if (type instanceof TypeName) {
+            return (TypeName) type;
+        }
+
+        if (type instanceof NonNullType) {
+            return this.extractName(((NonNullType) type).getType());
+        }
+
+        final String errorMessage = "Found unknown type: " + type.getClass().getSimpleName();
```
torbsto, line 94: What types are there? In the JavaDocs you say that it shouldn't parse ListType but I don't see an explicit check in this method for this case. I think that would be helpful instead of a generic error message.

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -59,4 +76,52 @@ public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveCo
     public boolean isValid(final TopicDirectiveContext context) {
         return context.getTopicDirective().hasKeyField() && !context.isListType();
     }
+
+    /**
+     * Extracts the {@link TypeName} of a  given {@link Type}. The class does not use the default implementation of the
+     * interface because the MutationRule should not parse ListType.
```
torbsto, line 82: Could you also add why it should not parse `ListType`?

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -59,4 +76,52 @@ public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveCo
...
+
+    private TypeName extractKeyFieldType(final TopicDirectiveContext context, final String keyField) {
+        final Optional<GraphqlElementParentTree> parentInfo = context.getEnvironment()
+            .getElementParentTree()
+            .getParentInfo();
+
+        if (parentInfo.isEmpty()) {
+            final String errorMessage = "Could not find the parent object type.";
+            log.error(errorMessage);
+            throw new QuickDirectiveException(errorMessage);
```
torbsto, line 107: See above; please don't log and throw

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/ModificationRule.java
```diff
@@ -42,14 +53,20 @@
...
+@Slf4j
 public class ModificationRule implements DataFetcherRule {
 
     @Override
     public List<DataFetcherSpecification> extractDataFetchers(final TopicDirectiveContext context) {
-        Objects.requireNonNull(context.getTopicDirective().getKeyField());
+        final String keyField = context.getTopicDirective().getKeyField();
+        Objects.requireNonNull(keyField);
+
+        final TypeName type = this.extractKeyFieldType(context, keyField);
```
torbsto, line 64: Could you explain why we need this? Just to check that it isn't a list? Why isn't it allowed to be a list?

raminqaf:
@torbsto I added some JavaDocs and explained everything

torbsto (approved):

PR #141, opened 2022-11-08
An extended user-part of the docs for the newly introduced --range-key option

### LINKED ISSUE #135: Add documentation for the range queries with `range-key`

DawidNiezgodka:
Resolves #135.


docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -268,6 +285,169 @@ Here you go - this is the list of the desired products.
...
+    price: Price
+    timestamp: Int
+}
+```
+As you can see, the field `productId` of the `Purchase` type
+has been removed, and `timestamp` has been added.
+The query has also changed.
+The new value of `keyArgument` - `id` - refers
+to the field you will mark
+with the `-range-key` option in the next step.
```
mkcode92, line 330:
```suggestion
with the `--range-key` option in the next step.  
```

docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -268,6 +285,169 @@ Here you go - this is the list of the desired products.
...
+type Purchase {
+    productId: Int!
+    userId: Int!
+    amount: Int
+    price: Price
+    timestamp: Int
+}
+```
+As you can see, the field `productId` of the `Purchase` type
+has been removed, and `timestamp` has been added.
```
mkcode92, line 326: The schemas look the same. Should they actually change?

torbsto (commented):
Maybe we could try to find a less negative wording :sweat_smile: Instead of seeing the original approach as a limitation, it is the baseline and the `--range-key` argument is a feature for more complex use cases.

docs/docs/user/getting-started/working-with-quick/range-query.md
DawidNiezgodka: Do you mean the `Purchase` type? 
They are not the same. In the [gateway part](https://bakdata.github.io/quick/0.8/user/getting-started/working-with-quick/gateway/) of the docs, the `Purchase` is:
```
type Purchase {
    purchaseId: String!
    productId: Int!
    userId: Int!
    amount: Int
    price: Price
}
```
and here:
```
type Purchase {
    productId: Int!
    userId: Int!
    amount: Int
    price: Price
    timestamp: Int
}
```
The `timestamp` has been added and the `purchaseId` is removed

docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -268,6 +285,169 @@ Here you go - this is the list of the desired products.
...
+  --header "content-type:application/json" \
+  --header "X-API-Key:$QUICK_API_KEY"\
+  --data "@./purchases.json"
+```
+Here is an example of the `purchases.json` file:
+??? "Example `purchases.json`"
+    ```
+    [
+        {
+          "key": "abc",
```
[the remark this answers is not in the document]
DawidNiezgodka: To implicitly show that contrary to "normal" range queries (w/o `--range-key`), there is no need to format data so that the message's key must appear as a field in the value.

docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -68,11 +68,26 @@ In our example, `timestampFrom` and `timestampTo` follow the naming scheme _fiel
...
-However, we suggest to follow this pattern to increase readability.
+However, we suggest following this pattern to increase readability.
 
 When you execute a range query, you receive a list of entries.
 Therefore, the return type of the query is a list of _UserRating_.
 
+!!! Important
+    With range queries as described here,
+    you can query data using the key of your Kafka messages,
+    i.e., the topic's key.
```
torbsto, line 79: [Try to avoid i.e.](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/ie)
```suggestion
    that is, the topic's key.
```

docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -68,11 +68,26 @@ In our example, `timestampFrom` and `timestampTo` follow the naming scheme _fiel
...
+    Thus, your data follows a specific format.
+    The value of the field chosen for `keyArgument`
+    must be the same as the value of the topic key.
+    [Below](#execute-the-query), 
+    you will notice that the data you send to the topic
+    follows this schema. The key of each message
```
torbsto, line 85:
```suggestion
    follows this schema. 
    The key of each message
```

docs/docs/user/getting-started/working-with-quick/range-query.md
```diff
@@ -268,6 +283,160 @@ Here you go - this is the list of the desired products.
...
+    productId: Int!
+    userId: Int!
+    amount: Int
+    price: Price
+    timestamp: Int
+}
+```
+As you can see, the `timestamp` has been added to the `Purchase` type.
+The query has also changed.
+The new value of `keyArgument` - `id` - refers
```
torbsto, line 320:
```suggestion
The new value of `keyArgument` - `userId` - refers
```

torbsto (changes requested):

torbsto (approved):

PR #147, opened 2022-11-15
e2e test for the range-key option

### LINKED ISSUE #134: Create E2E test for range queries with `range-key`
We should add E2E test for the new `range-key` functionality

### LINKED ISSUE #145: Fix typo in the range-key docs
There is a typo in the docs about the range key, namely here:
```
query {
    findUserPurchasesInTime(id: "2", timestampFrom: 1, timestampTo: 4) {
    productId,
    price
    {
        total
    }
  }
}
```

 It should be `userId` and not `id`

DawidNiezgodka:
added an e2e test for range-key, updated the justfile. resolves #145 and resolves #134

### DIFF
--- a/e2e/functional/key-range/result-range.json
+++ b/e2e/functional/key-range/result-range.json
@@ -1,0 +1,23 @@
+[
+  {
+    "productId": 111,
+    "price": {
+      "total": 14.99
+    },
+    "timestamp": 1
+  },
+  {
+    "productId": 111,
+    "price": {
+      "total": 19.99
+    },
+    "timestamp": 2
+  },
+  {
+    "productId": 111,
+    "price": {
+      "total": 24.99
+    },
+    "timestamp": 3
+  }
+]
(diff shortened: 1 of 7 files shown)


e2e/functional/key-range/result-range.json
```diff
@@ -0,0 +1,23 @@
+[
```
torbsto, line 1: Those aren't the expected results, right? The `productId` is the same for each and the totals aren't even in the input data.

DawidNiezgodka: ah, ye, sry, I took the result from the docs but copied the one that relates to range-query and not range-query with key option

e2e/functional/key-range/result-range.json
```diff
@@ -0,0 +1,14 @@
...
+      {
+        "productId": 123,
+        "price": {
+          "total": 30
```
torbsto, line 5: Does this work? I'd assume it must be `30.00`

DawidNiezgodka: I believe I copied it directly from Altair and not wrote it by hand so it should function. I will check it later as I'm having some problems creating topics atm.

torbsto (approved):

PR #150, opened 2022-11-30
keyArgument type sets the key serializer

### LINKED ISSUE #120: Gateway key serializer should be set by the keyArgument type
In a Gateway, it is important to create a partitioned-aware mirror client based on the newly defined key. Currently, we are using the information in the topic registry (i.e., key serde) to serialize the keys and find the partition. This should change to the type of the new key. One idea is to use the type of `keyArgument` and supply the SerDe.

This issue is related to https://github.com/bakdata/quick/issues/114.

raminqaf:
Closes #120


gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/RangeFetcherRule.java
```diff
@@ -80,4 +91,32 @@ public boolean isValid(final TopicDirectiveContext context) {
...
+
+        if (parentInfo.isEmpty()) {
+            throw new QuickDirectiveException("Could not find the parent object type.");
+        }
+
+        final GraphQLSchemaElement element = parentInfo.get().getElement();
+        final ObjectTypeDefinition definition = ((GraphQLObjectType) element).getDefinition();
+
+        final List<InputValueDefinition> inputValueDefinitions =
+            definition.getFieldDefinitions().get(0).getInputValueDefinitions();
```
torbsto, line 108: Why `get(0)`?

common/src/main/java/com/bakdata/quick/common/api/client/mirror/DefaultMirrorClientFactory.java
```diff
@@ -35,4 +37,14 @@ public <K, V> MirrorClient<K, V> createMirrorClient(final HttpClient client, fin
             new MirrorValueParser<>(valueTypeResolver, client.objectMapper());
         return new DefaultMirrorClient<>(mirrorHost, mirrorValueParser, requestManager);
     }
+
+    @Override
+    public <K, V> MirrorClient<K, V> createMirrorClient(final HttpClient client, final String topic,
+        final Serde<K> keySerde, final TypeResolver<V> valueTypeResolver) {
```
torbsto, line 43: Why do we need the keySerde here? It doesn't seem to be used (?)

torbsto (commented):
This seems like quite a small change to me. Do you have a concrete alternative in mind that would be more elegant?

gateway/src/main/java/com/bakdata/quick/gateway/directives/topic/rule/fetcher/RangeFetcherRule.java
raminqaf: I found a more elegant way to retrieve the keyArgument type 🙂

common/src/main/java/com/bakdata/quick/common/api/client/mirror/MirrorClientFactory.java
```diff
@@ -17,14 +17,35 @@
...
+     * @param quickTopicData The quick topic data
+     * @param <K> Type of the key
+     * @param <V> Type of the value
+     * @return A {@link MirrorClient}
+     */
+    default <K, V> MirrorClient<K, V> createMirrorClient(final HttpClient client, final String topic,
+        final Lazy<QuickTopicData<K, V>> quickTopicData) {
+        final MirrorHost mirrorHost = MirrorHost.createWithPrefix(topic);
+        final MirrorRequestManager requestManager = new DefaultMirrorRequestManager(client);
+        final TypeResolver<V> valueTypeResolver = quickTopicData.get().getValueData().getResolver();
```
torbsto, line 43: Did you consider what happens if the gateway and its schema are created before the topics? This is the reason why `Lazy` exist in the first place.

torbsto (approved):

PR #151, opened 2022-11-30
Replace deprecated set-output command in GitHub actions

### LINKED ISSUE #109: Replace deprecated `set-output` commands
Warning: The set-output command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information, see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

raminqaf:
closes #109

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,9 +42,9 @@ jobs:
         id: read-config
         # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
         run: |
-          CONTENT=$(cat ./.github/changelog-config.json)
-          echo "CONTENT<<EOF" >> $GITHUB_OUTPUT
-          echo "$CONTENT" >> $GITHUB_OUTPUT
+          content=$(cat ./.github/changelog-config.json)
+          echo "content<<EOF" >> $GITHUB_OUTPUT
+          echo "$content" >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -35,14 +35,17 @@ jobs:
...
-          echo "::set-output name=tag::$tag"
+          echo "oldTag=$(just print-version)" >> $GITHUB_OUTPUT
+          echo "tag=$(just bump-version ${{ github.event.inputs.scope }})" >> $GITHUB_OUTPUT
 
       - name: Read changelog config
         id: read-config
-        run: echo ::set-output name=content::$(cat ./.github/changelog-config.json)
+        # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+        run: |
+          CONTENT=$(cat ./.github/changelog-config.json)
```
torbsto, line 45: Is this case-sensitive? We reference it by ` ${{ steps.read-config.outputs.content }}`

raminqaf: No, it's not. If you take a look at the quick-cli, you will notice that we have the same there:
https://github.com/bakdata/quick-cli/blob/04384658353c3be8e49fa6d7fb42300c1befcd32/.github/workflows/release.yaml#L63
I can lowercase it if it is confusing.

torbsto: I think it would be a bit easier to understand, but it isn't a blocker for me

torbsto (approved):

PR #153, opened 2022-12-01
Fix mirror deletion on range key

### LINKED ISSUE #155: Fix deletion of Mirrors with `--range-key` option
Currently, the deletion of Mirrors having the `--range-key` option set doesn't work correctly. During the clean-up job, we need the schema from the topic registry to build the topology. However, we are deleting the topic along with the schema from the topic registry before running the clean-up job.
The solution would be to move the deletion of the topic from the topic registry after the cleanup job is done.

raminqaf:
Closes #155


torbsto (commented):
> When the cleanup job started, it was trying to do the repartitioning. This should be skipped if the cleanup flag is set.

I don't see why it should be skipped. The repartition topic should be cleaned like any other topic of the topology.

What is the error you're trying to fix?

raminqaf:
> > When the cleanup job started, it was trying to do the repartitioning. This should be skipped if the cleanup flag is set.
> 
> I don't see why it should be skipped. The repartition topic should be cleaned like any other topic of the topology.
> 
> What is the error you're trying to fix?

The reason I do the check and skip it is that whenever the cleanUp job starts, the `ParsedSchema` is null therefore, we cannot create a repartitioned stream.
https://github.com/bakdata/quick/blob/83055ee46ee45a0b2f0b1a5ba1760184b4c7ade7/mirror/src/main/java/com/bakdata/quick/mirror/IndexInputStreamBuilder.java#L102-L113

torbsto:
Then you somehow have to delay deleting the subject in the schema registry until the clean up job is done. One option would be to move the deletion into the mirror.

torbsto (approved):

